PR #9, opened 2017-04-12
Remove absolute path to coap-client

rubenbe:
Most linux distribution (including the dev Docker container) have /usr/local/bin configured in their PATH.
Plus removing the absolute path also allows users to have coap-client installed in a different location without having to update the source code to reflect this change.


balloob:
Thanks!

Issue #2, opened 2017-04-11
[Research] Support for observing/subscribe/streaming?


lwis:
`aiocoap` has plenty of examples, one of them being;

https://github.com/chrysn/aiocoap/blob/master/clientGET-observe.py

kmjonsson:
The problem with aiocoap is the lack of DTLS-support. See: https://github.com/chrysn/aiocoap/issues/67

The coap-client has observe support as well (-s). They call it subscribe but as far as I understand from the specification this is the same thing.

lwis:
@kmjonsson Good point. I'm going to try and get hold of some bulbs and a gateway to test out observing, but it may be a while so if anyone else can confirm it's working then we can start development.

kmjonsson:
@lwis, yes it works. It will leave the coap-client running and pop out new messages if there are changes. This also includes the switches so they can be used within HA (or other) as triggers for any event.
I guess it also includes groups, and other types of GET commands like /.well-known/core etc..

Popping out a observing/subscribing thread for each bulb/switch/thing might be a little silly but until there are a native coap library with dtls support this might be the best thing.

If you need som sample output I can provide it but I'm currently unavailable for doing coding but I will try to do some testing.

balloob:
The app is able to do it and I doubt that it would open an observe channel for each bulb/group but I haven't explored the API that much beyond the basics yet.

kmjonsson:
@balloob, it looks like the COAP observe/subscribe works like that. But the app only needs to observe things that are currently "active".

balloob:
@kmjonsson you're right! Adding `-s <duration>`, ie `-s 50` and point it at a light allowed me to stream updates. This is awesome 🎉

balloob:
Sadly, it still allows me to only stream one device at a time. There was a notifications endpoint `15006` but it only reports on notifications like loss of internet connectivity.

balloob:
It means that we need to spin up X processes, one for each light 😞

lwis:
Damn, a little more feasible if we had aiocoap - but that's not ready yet.

rubenbe:
@balloob I guess a solution would be to integrate libcoap.so directly into python using e.g. cffi. As such we can support more complex use-cases (like listening to multiple devices) without having to spawn a process for each.

lwis:
@rubenbe I'm of the opinion that those efforts would be better spent enhancing the incompatibilities with aiocoap (progress seems to have stalled)

[...]

PR #12, opened 2017-04-17
Disable coap-client debug output.

rubenbe:
Prevent coap-client from outputting annoying debug output to stdout, which we need to filter out afterwards.
Using some autoconf magic ensures no modifications to the coap source code are made.
This might also be useful for #2, as coap-client now provides a clean json stream when subscribed.

### DIFF
--- a/pytradfri/coap_cli.py
+++ b/pytradfri/coap_cli.py
@@ -35,7 +35,7 @@ def api_factory(host, security_code):
 
         kwargs = {
             'timeout': timeout,
-            'stderr': subprocess.STDOUT,
+            'stderr': subprocess.DEVNULL,
         }
 
         if data is not None:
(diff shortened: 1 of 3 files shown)


pytradfri/coap_cli.py
```diff
@@ -48,21 +48,13 @@ def request(method, path, data=None, *, parse_json=True, timeout=10):
...
             raise RequestTimeout() from None
         except subprocess.CalledProcessError as err:
             raise RequestError(
                 'Error executing request: %s'.format(err)) from None
 
-        # Return only the last line, where there's JSON
-        lines = [line for line in out.split('\n')[1:]
-                 if 'decrypt_verify' not in line]
-
-        if not lines:
```
balloob: We will still need this check as some commands do not give a return value.

balloob:
Fixed it myself. Thanks for the contribution 👍

PR #15, opened 2017-04-18
Removed identity since it's not needed

ws141:
I noticed that the identity (-u user) property is not needed when using the coap-client so I removed it.

In fact it seems advised to actually not use it according to [this article](https://bitsex.net/software/2017/coap-endpoints-on-ikea-tradfri/)
 >   UPDATE: Leave PSK identity blank.

### DIFF
--- a/pytradfri/coap_cli.py
+++ b/pytradfri/coap_cli.py
@@ -22,8 +22,6 @@ def api_factory(host, security_code):
 
         command = [
             'coap-client',
-            '-u',
-            'Client_identity',
             '-k',
             security_code,
             '-v',


balloob:
Confirmed that it works, thanks 🐬

balloob:
Interestingly, if I use the OS X build provided by [`homebridge-tradfri`](https://github.com/jesper-lindberg/homebridge-tradfri/tree/master/bin) I am required to add `-u Client_identity`. Going to revert this change for now until we figure out what is happening.

balloob:
Nevermind, that build was old. I made a build using our own instructions and it all works.

PR #16, opened 2017-04-22
Added new constants

ggravlingen:
These are required when adding support for automations (wake up, not at home and lights off.)


ggravlingen:
The constants are included in the smask task PR.

PR #18, opened 2017-04-23
Add smart task

ggravlingen:
This will add basic support to access tasks and get information on them. For now, it is not possible to add, remove or manipulate the tasks.


pytradfri/smart_task.py
```diff
@@ -0,0 +1,242 @@
...
+
+    def __getattr__(self, attr):
+        try:
+            return self._lookup[attr]
+        except KeyError:
+            raise AttributeError(attr)
+
+    def get_selected_keys(self, selection):
+        """Return a list of keys for the given selection."""
+        return [k for k, b in self._lookup.iteritems() if b & selection]
```
MartinHjelmare: `dict.iteritems` is python 2 only.

pytradfri/smart_task.py
```diff
@@ -0,0 +1,242 @@
...
+
+    @property
+    def repeat_days(self):
+        """Return int (bit) for enabled weekdays."""
+        return self.raw.get(ATTR_REPEAT_DAYS)
+
+    @property
+    def repeat_days_list(self):
+        """Binary representation of weekdays the event takes place."""
+        print(WEEKDAYS.get_selected_values(self.raw.get(ATTR_REPEAT_DAYS)))
```
MartinHjelmare: `print`? It should be `return` right?

ggravlingen:
Thanks @MartinHjelmare for reviewing!

pytradfri/smart_task.py
ggravlingen: Yes, forgot to remove it after doing some debuggning. 😊

pytradfri/smart_task.py
ggravlingen: According to some page I read, dict.items is fine in Python 3 though? Changed to use that instead.

MartinHjelmare: Yes. `dict.items` works both in python 2 and 3, but is more efficient in python 3. `dict.iteritems` is more efficient in python 2 than `dict.items`.

ggravlingen: Thanks!

pytradfri/gateway.py
```diff
@@ -65,6 +67,17 @@ def _get_mood_parent(self):
         """Get the parent of all moods."""
         return self.api('get', [ROOT_MOODS])[0]
 
+    def get_smart_tasks(self):
+        """Return the transitions linked to the gateway."""
+        tasks = self.api('get', [ROOT_SMART_TASKS])
+
+        return [self.get_task(task) for task in tasks]
+
+    def get_task(self, task_id):
```
balloob: We should be consistent in the naming, this one should either be called `get_smart_task` or the other one be `get_tasks`

pytradfri/smart_task.py
```diff
@@ -0,0 +1,242 @@
...
+
+    @property
+    def repeat_days_list(self):
+        """Binary representation of weekdays the event takes place."""
+        return WEEKDAYS.get_selected_values(self.raw.get(ATTR_REPEAT_DAYS))
+
+    @property
+    def start_action(self):
+        """Return state and all devices in smart action."""
+        return self.raw.get(ATTR_START_ACTION)
```
balloob: Shouldn't this be represented as a class?

pytradfri/smart_task.py
```diff
@@ -0,0 +1,242 @@
...
+
+    @property
+    def start_action_state(self):
```
balloob: We should remove this property, people can access it via `task.start_action.state`

pytradfri/smart_task.py
ggravlingen: You are right. I'll make that change in my next PR.

pytradfri/smart_task.py
ggravlingen: I'll add it as a method when I att the start_action-class.

pytradfri/gateway.py
ggravlingen: This has been resolved.

PR #19, opened 2017-04-23
Reorder coap_client arguments. Fixes #14.

viernullvier:
This is a cleaner version of @diplix' patch (#17).

PR #20, opened 2017-04-24
Add support for observing resources.

balloob:
This adds support for observing devices. I did shuffle some arguments around (related to #17 and #19). I think that the problem is related to having the arguments be after the url.

Returned API object now has a new method `observe`. Call it with a path, time to observe and a callback.

I've added support to this for devices, so that you can observe a device, which will update and then call your callback.

```python
def beer(device):
  print(device.name + " is now " + str(device.light_control.lights[0].state))

lights[0].observe(beer)
```

**Update:** generalized the code so that now we can observe groups and moods too.
**Update 2:** Added `set_dimmer` and `set_state` to the group object.

### DIFF
--- a/pytradfri/coap_cli.py
+++ b/pytradfri/coap_cli.py
@@ -2,6 +2,7 @@
 import json
 import logging
 import subprocess
+from time import time
 
 from .error import RequestError, ClientError, ServerError, RequestTimeout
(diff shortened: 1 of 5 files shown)


lwis:
Resolves #2.

PR #27, opened 2017-04-29
Added support for transition time

demianhiemstra:


pytradfri/device.py
```diff
@@ -167,6 +170,13 @@ def set_xy_color(self, color_x, color_y, *, index=0):
             ATTR_LIGHT_COLOR_Y: color_y
         }, index=index)
 
+    def set_transition_time(self, transition_time):
```
balloob: We should not store transition time as a transient property on an instance. Instead, make it an optional keyword argument to `set_dimmer`:

```python
def set_dimmer(self, dimmer, *, index=0, transition_time=None):
    values = {
        ATTR_LIGHT_DIMMER: dimmer,
    }
    if transition_time is not None:
        values[ATTR_TRANSITION_TIME] = transition_time
```

balloob:
Awesome, thanks! 🐬

PR #30, opened 2017-05-03
Smart tasks

ggravlingen:
This PR adds support for Smart Tasks, a built-in functionality to configure starting lights on certain time-intervals. Not all features are currently working as initial focus has been on the "Wake up" smart task. This class can set transition time and dimmer level for a specific device in a task, which is currently not possible in the app.

Do not merge, am working on building proper test cases.

### DIFF
--- a/pytradfri/smart_task.py
+++ b/pytradfri/smart_task.py
@@ -94,10 +94,7 @@ class SmartTask(ApiResource):
 
     @property
     def path(self):
-        """Return gateway path.
-
-        Migrate to ApiResource when possible.
-        """
+        """Return gateway path."""
         return [ROOT_SMART_TASKS, self.id]
 
     @property
--- a/pytradfri/smart_task.py
+++ b/pytradfri/smart_task.py
@@ -175,7 +175,7 @@ class SmartTask(ApiResource):
     @property
     def start_action(self):
         """Return start action object."""
-        return StartAction(self, self.api, self.path, self._gateway)
+        return StartAction(self, self.api, self.path)
 
     def __repr__(self):
         """Return a readable name for smart task."""
@@ -203,7 +203,8 @@ class TaskControl:
             i,
             self.state,
             self.api,
-            self.path) for i in range(len(self.raw))]
+            self.path,
+            self.raw) for i in range(len(self.raw))]
 
     def set_dimmer_start_time(self, hour, minute):
         """Set start time for task (hh:mm) in iso8601.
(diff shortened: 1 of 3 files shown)


pytradfri/smart_task.py
```diff
@@ -183,51 +171,157 @@ def task_start_time_seconds(self):
...
+        #  This is to calculate the difference between local time
+        #  and the time in the gateway
+        d1 = dt.strptime(
+                self._gateway.get_gateway_info().current_time_iso8601,
+                '%Y-%m-%dT%H:%M:%S.%fZ')
+        d2 = dt.strptime(
+                dt.now().isoformat(),
+                '%Y-%m-%dT%H:%M:%S.%f')
+
+        diffMinutes = round(int(
```
balloob:
```python
diff = d1 - d2  # this is a time delta object that you can add/subtract to dates

# Example
d4 = d3 + diff
```

balloob: Oh, and snake_case over camelCase is the Python way 👍

pytradfri/smart_task.py
```diff
@@ -183,51 +171,157 @@ def task_start_time_seconds(self):
...
+        command = {
+            ATTR_SMART_TASK_TRIGGER_TIME_INTERVAL:
+                [{
+                    ATTR_SMART_TASK_TRIGGER_TIME_START_HOUR: newtime.hour,
+                    ATTR_SMART_TASK_TRIGGER_TIME_START_MIN: newtime.minute
+                }]
+            }
+        self.set_values(command)
+
+    def set_values(self, command):
```
balloob: Why not remove this and call directly `self._task.set_values(command)` above ?

pytradfri/smart_task.py
```diff
@@ -183,51 +171,157 @@ def task_start_time_seconds(self):
...
         """Initialize TaskInfo."""
         self.task = task
         self.index = index
+        self.state = state
+        self.api = api
+        self.path = path
+        self._raw = raw
+
+    @property
+    def devices_json_list(self):
```
balloob: We don't have JSON here, we just have Python objects. Let's remove `json` from the name.

Also, the name list is a bit weird because you're returning a dictionary.

pytradfri/smart_task.py
```diff
@@ -183,51 +171,157 @@ def task_start_time_seconds(self):
     @property
```
ggravlingen: I've renamed the method and am now returning a time-object.

balloob:
Should be good to go when the tests pass 👍

PR #32, opened 2017-05-15
Split coap-client output on newline.

slaveriq:
With coap-client v4.1.2  the output can not be blindly parsed as json.
Example output:
```
v:1 t:CON c:GET i:6697 {} [ ]
{}
```
Splitting on newline and using the 2nd part/line seems to solve that issue.

### DIFF
--- a/pytradfri/coap_cli.py
+++ b/pytradfri/coap_cli.py
@@ -128,5 +128,7 @@ def _process_output(output, parse_json=True):
 
     elif not parse_json:
         return output
-
+    if len(output.split('\n')) == 2:
+        output = output.split('\n')[1]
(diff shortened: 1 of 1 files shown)


balloob:
The recommended approach is to use our install script which is not on 4.1.2 https://github.com/ggravlingen/pytradfri/blob/master/script/install-coap-client.sh

slaveriq:
I get the same output from that version.

balloob:
With that version, the debug output is printed to stderr. So it will work with pytradfri.

slaveriq:
Alright, but why not make it work with upstream?

ggravlingen:
I'm closing this based on the recommendation from @balloob on how to make it work above.

PR #33, opened 2017-05-29
Support for wrapping API requests

lwis:
If you wrap the func, you can't get access to `observe`. I was debating on whether or not this should have been a class?

### DIFF
--- a/pytradfri/api.py
+++ b/pytradfri/api.py
@@ -1,10 +1,12 @@
 """API utilities."""
+from functools import wraps
 
 from .error import RequestTimeout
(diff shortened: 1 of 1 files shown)


balloob:
This is the wrong approach. The coap_cli class should not be aware of any wrapper to keep things simple. Instead, we should just assign `observe` in the `retry_timeout` func to achieve the same.

lwis:
@balloob I woke up and realised we just need to use `functools.wraps`.

balloob:
#Winning

PR #34, opened 2017-05-30
Refactor for command pattern to become I/O independent

lwis:
Hopefully this is on the right lines, feedback/discussion welcome!


pytradfri/parser.py
```diff
@@ -0,0 +1,35 @@
...
+    if not output:
+        return None
+
+    elif 'decrypt_verify' in output:
+        raise RequestError(
+            'Please compile coap-client without debug output. See '
+            'instructions at '
+            'https://github.com/ggravlingen/pytradfri#installation')
+
+    elif output.startswith(CLIENT_ERROR_PREFIX):
```
balloob: I wonder if aiocoap has a different way of notifying us of these errors

pytradfri/smart_task.py
```diff
@@ -11,6 +11,8 @@
 
 from datetime import (datetime as dt)
 import datetime
+
+from pytradfri.command import Command
```
balloob: Probably want to be consistent and stick with relative imports.

balloob:
I like it ! 

This is pretty much what I envisioned honestly, awesome job ! 👍

pytradfri/aiocoap_api.py
```diff
@@ -0,0 +1,88 @@
...
+from pytradfri.command import Command
+from .parser import process_output
+
+_LOGGER = logging.getLogger(__name__)
+
+
+def api_factory(host, security_code):
+    """Generate a request method."""
+
+    PSK_STORE[b'Client_identity'] = security_code.encode('utf-8')
```
balloob: So this is how to auth using my fork. The real branch has evolved a bit more and we should make sure we fix the auth against that before merging.

Security is set as a property on the transport: https://github.com/chrysn/aiocoap/blob/tinydtls/aiocoap/transports/tinydtls.py#L182

pytradfri/aiocoap_api.py
```diff
@@ -0,0 +1,88 @@
...
+            _LOGGER.debug('Executing %s %s %s', host, method, path)
+
+        api_method = GET
+        if method == 'put':
+            api_method = PUT
+
+        protocol = yield from Context.create_client_context()
+        msg = Message(code=api_method, uri=url, **kwargs)
+        res = yield from protocol.request(msg).response
+        _LOGGER.debug("RECEIVED STATUS", res.code)
```
balloob: Based on res.code we can raise ClientError / ServerError ?

pytradfri/aiocoap_api.py
```diff
@@ -0,0 +1,88 @@
...
+        _LOGGER.debug("RECEIVED PAYLOAD", res.payload.decode('utf-8'))
+
+        api_command.result = process_output(res, parse_json)
```
balloob: I wonder if `process_output` can be made a method on Command: `process_response`. We could even pass `res.code` in optionally, allowing process_response to raise / forward an exception to the callback.

pytradfri/aiocoap_api.py
lwis: I've rebased my branch against the upstream tinydtls branch, but I'm confused about a few things;

https://github.com/chrysn/aiocoap/blob/tinydtls/aiocoap/transports/tinydtls.py#L56 - is this a temporary hack?

https://github.com/chrysn/aiocoap/blob/tinydtls/aiocoap/protocol.py#L525 - this and DTLSSecurityStore are created inline, and the only way to access the transport_protocols is from the context?

I can't see a clean way of providing auth details to the context so they're passed down to the DTLSSecurityStore?

balloob: I think that for now we can just monkey patch the store. It's not clear to me how else it can be used.

We should probably ask the author.

pytradfri/aiocoap_api.py
lwis: I took another look at `process_output` and the only similarity will likely end up being parsing the JSON, unless I've missed something, it looks like the aiocoap representation of codes is different?

Raising exceptions in callbacks always feels messy to me. We could lazily process the response, raising an exception when the result is requested? Although, I can't visualise this using the command pattern for requests such as `PUT`.

pytradfri/aiocoap_api.py
lwis: Done.

lwis:
Pushed another commit with some working examples.

NB: To get the examples to work I had to locally edit;
https://github.com/chrysn/aiocoap/blob/tinydtls/aiocoap/transports/tinydtls.py#L171 

Because this takes 3 args...
https://git.fslab.de/jkonra2m/tinydtls/blob/cython/cython/dtls.pyx#L258

I believe the examples also serve as a good usage example. For example, returning a list of commands for the list of devices, which the user then has to iterate over themselves. I've watched the PyCon talk, and agree that we shouldn't be doing the IO ourselves, but does that look as expected?

pytradfri/command.py
```diff
@@ -0,0 +1,70 @@
...
+        return self._raw_result
+
+    @property
+    def result(self):
+        return self._result
+
+    @result.setter
+    def result(self, value):
+        if self._callback:
+            self._result = self._callback(value)
```
balloob: I think that callback is the wrong name here. Callback to me indicates a side effect but instead we're using it to also process the result.

But then again, not sure what would be a good name either.

balloob: I think that the main purpose is to process the result, with optionally a side effect in case it's an update call. Let's go for `process_result` ?

setup.py
```diff
@@ -12,5 +17,6 @@
   author='balloob, ggravlingen',
   author_email='no@email.com',
   url='https://github.com/ggravlingen/pytradfri',
-  download_url=DOWNLOAD_URL
+  download_url=DOWNLOAD_URL,
+  install_requires=REQUIRES,
```
balloob: We should mark these as optional requires. 

```
extras_require={'asyncio': REQUIRES},
```

balloob: Since we depend on a branch of aiocoap to be installed, should we actually depend on it in our setup.py ? We wouldn't want that version to be installed right ?

pytradfri/aiocoap_api.py
```diff
@@ -0,0 +1,136 @@
...
+            try:
+                res = yield from type(it).__anext__(it)
+            except StopAsyncIteration:  # TODO: This was added in 3.5 :(
+                running = False
+            else:
+                result = _process_output(res)
+                callback(result)
+
+    # This will cause a RequestError to be raised if credentials invalid
+    request(Command('get', ['status']))
```
balloob: you'll need to yield from this.

lwis: Good spot!

example_async.py
```diff
@@ -0,0 +1,108 @@
...
+    devices_commands = devices_command.result
+    devices = []
+    for device_command in devices_commands:
+        yield from api(device_command)
+        devices.append(device_command.result)
+
+    lights = [dev for dev in devices if dev.has_light_control]
+
+    tasks_command = gateway.get_smart_tasks()
+    yield from api(tasks_command)
```
balloob: yielding from the api will actually return the result, no need to fetch it from the command afterwards.

setup.py
lwis: Yup, this was something on my list to Google.

I was also considering moving the aiocoap imports to the func, as they'll be imported as part of the __main__.py loading.

balloob: that's a good idea.

balloob: Actually, let's instead no longer import the factories into `__init__.py` but instead have users import from `pytradfri.aiocoap` etc.

ggravlingen:
I'm following this with interest. Great work guys! :+1: :1st_place_medal:

balloob:
Don't forget the tests @lwis

lwis:
@balloob done, I've got rid of the mock_api for now - I felt like the unit tests were integration tests, as I feel it would be more correct for us to assert the returned command - rather than testing that a mock_api can execute with our produced command.

lwis:
I think this is ready for testing!

balloob:
Great work! Will play with it soon.

ggravlingen:
How can this be tested? Just download the branch and run commands from CLI as usual?

lwis:
@ggravlingen example files are a good reference.

pytradfri/api/aiocoap_api.py
```diff
@@ -82,17 +80,19 @@ def _execute(protocol, api_command):
             api_method = Code.PUT
 
         msg = Message(code=api_method, uri=url, **kwargs)
-        with timeout(request_timeout):
```
balloob: Why would you not use `async_timeout` ?

lwis: They have an internal mechanism for request timeouts, it seemed unnecessary.

pytradfri/api/aiocoap_api.py
```diff
@@ -48,10 +48,13 @@ def api_factory(host, security_code):
...
     PatchedDTLSSecurityStore.SECRET_PSK = security_code
 
-    protocol = yield from Context.create_client_context()
-    for endpoint in protocol.transport_endpoints:
-        if type(endpoint) == TransportEndpointTinyDTLS:
-            pass
+    @asyncio.coroutine
+    def _get_protocol():
+        protocol = yield from Context.create_client_context()
+        for endpoint in protocol.transport_endpoints:
```
MartinHjelmare: What happens here? Maybe add a comment?

balloob:
The Dockerfile seems to be broken so I can't test it :(

lwis:
@balloob Just rebuilt and re-ran, all working for me. What issue are you seeing exactly?

pytradfri/api/aiocoap_api.py
```diff
@@ -0,0 +1,162 @@
+"""Coap implementation using aiocoap."""
+import asyncio
+import json
+import logging
+import sys
+
+sys.path.insert(0, '/usr/src/build/tinydtls/cython')  # noqa
```
balloob: This is not something that we should solve inside the lib. Instead we should provide instructions for people to make sure that we have access to the right dependencies.

pytradfri/api/aiocoap_api.py
```diff
@@ -0,0 +1,162 @@
...
+    """Generate a request method."""
+
+    security_code = security_code.encode('utf-8')
+
+    PatchedDTLSSecurityStore.SECRET_PSK = security_code
+
+    @asyncio.coroutine
+    def _get_protocol():
+        protocol = yield from Context.create_client_context()
+        for endpoint in protocol.transport_endpoints:
```
balloob: Agree with @MartinHjelmare, please add a comment what this code does. Now it just looks to me like code that does nothing.

pytradfri/api/aiocoap_api.py
```diff
@@ -0,0 +1,162 @@
...
+def _patched_datagram_received(self, data, addr):
+    self._dtls_socket.handleMessage(self._connection, data, 0)
+
+
+aiocoap.transports.tinydtls.DTLSClientConnection.datagram_received = \
+    _patched_datagram_received
+
+
+@asyncio.coroutine
+def api_factory(host, security_code):
```
balloob: We should allow passing in the event loop as an optional keyword

pytradfri/api/aiocoap_api.py
```diff
@@ -0,0 +1,162 @@
...
+    @asyncio.coroutine
+    def request(*api_commands):
+        """Make a request."""
+        if len(api_commands) == 1:
+            result = yield from _execute(api_commands[0])
+            return result
+
+        command_results = []
+
+        for api_command in api_commands:
```
balloob: Is there any reason why we would do this in sequence instead of in parallel?

```python
command_results = yield from asyncio.gather(
    *(_execute(api_command) for api_command in api_commands), loop=loop)
```

(this will require to pass `loop` into api_factory and default to `loop = asyncio.get_event_loop()`)

pytradfri/api/libcoap_api.py
```diff
@@ -60,11 +68,31 @@ def request(method, path, data=None, *, parse_json=True, timeout=10):
...
+        api_command.result = _process_output(return_value, parse_json)
+        return api_command.result
+
+    def request(*api_commands):
+        """Make a request."""
+        if len(api_commands) == 1:
+            return _execute(api_commands[0])
 
-    def observe(path, callback, duration):
+        command_results = []
```
balloob: Optional: you could use a string comprehension here 🎉

lwis: String comp? What would that look like?

pytradfri/api/aiocoap_api.py
```diff
@@ -41,7 +38,7 @@ def _patched_datagram_received(self, data, addr):
 
 
 @asyncio.coroutine
-def api_factory(host, security_code):
+def api_factory(host, security_code, loop=asyncio.get_event_loop()):
```
balloob: You should default it to `None` because the call to `get_event_loop` will be cached

balloob:
I pushed a small update to the async example to simplify the observe code slightly 9ea2f4c30e62d46518d0f314e361b5bb3f968f85

balloob:
@lwis I love it. I think this is almost ready to ship. Last step is to make sure we have some better documentation around how to get tinydtls built.

 - Add a comment why `sys.path.insert` statement is needed and what it points to in example_async
 - Update README:
    - Installation instructions for aiocoap (all installation instructions should use our `sh` build scripts)
    - add yourself to acknowledgements!

For future PR
- Update `install-aiocoap.sh` to use same flags for tinydtls as `install-coap-client.sh` does, so that we don't print debug stuff to the console.

lwis:
I think I got it all, also updated the main, haven't tested that yet, though.

balloob:
I tested it.

 - The transport imports broke somehow, fixed it by importing module directly.
 - Making requests at same time blows up gateway. For now revert back to do it in sequence.
 - Simplify the installation of tinydtls by installing it as a Python package

lwis:
:+1: looking good! Was about to ask why you reverted the gather.

balloob:
I _think_ this is ready to merge. What do you think @lwis? @ggravlingen, any comments?

lwis:
Yup, super happy with this now. I think if we continue tweaking, we'll never stop 😉

PR #37, opened 2017-06-15
Cache protocol for better performance, and to fix observation issues

lwis:
Feedback welcome, if there's a better way of handling this please let me know.

The reasons for this;

- If a new protocol is created for every request, a maximum of 3 observations can be obtained, this may be due to sockets or limitations elsewhere.
- In my testing, if a protocol is created and the Gateway is turned off, then turned on again, no further requests appear to be successful

### DIFF
--- a/pytradfri/api/aiocoap_api.py
+++ b/pytradfri/api/aiocoap_api.py
@@ -58,6 +58,7 @@ def api_factory(host, security_code, loop=None):
         """Reset the protocol if an error occurs.
            This can be removed when chrysn/aiocoap#79 is closed."""
         # Be responsible and clean up.
+        protocol = yield from _get_protocol()
         yield from protocol.shutdown()
         nonlocal _protocol
         _protocol = None
--- a/pytradfri/api/aiocoap_api.py
+++ b/pytradfri/api/aiocoap_api.py
@@ -14,8 +14,6 @@ from ..command import Command
 
 _LOGGER = logging.getLogger(__name__)
 
-aiocoap.numbers.constants.MAX_RETRANSMIT = 10
-
 
 class PatchedDTLSSecurityStore:
     """Patched DTLS store in lieu of impl."""
@@ -45,7 +43,8 @@ def api_factory(host, security_code, loop=None):
 
     PatchedDTLSSecurityStore.SECRET_PSK = security_code.encode('utf-8')
 
-    _protocol = None
+    _observations_err_callbacks = []
+    _protocol = yield from Context.create_client_context(loop=loop)
 
     @asyncio.coroutine
     def _get_protocol():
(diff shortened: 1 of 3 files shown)


pytradfri/api/aiocoap_api.py
```diff
@@ -43,15 +43,43 @@ def api_factory(host, security_code, loop=None):
...
+    _protocol = None
 
     @asyncio.coroutine
     def _get_protocol():
         """Get the protocol for the request."""
-        protocol = yield from Context.create_client_context(loop=loop)
-        return protocol
+        nonlocal _protocol
+        if not _protocol:
+            _protocol = yield from Context.create_client_context(loop=loop)
```
balloob: Why not create a client context as part of the api_factory initialization ? I'm assuming that it doesn't do any I/O ?

pytradfri/api/aiocoap_api.py
```diff
@@ -43,15 +43,43 @@ def api_factory(host, security_code, loop=None):
...
+        _protocol = None
+
+    @asyncio.coroutine
+    def _get_response(msg):
+        """Perform the request, get the response."""
+        try:
+            protocol = yield from _get_protocol()
+            pr = protocol.request(msg)
+            r = yield from pr.response
+            return (pr, r)
```
balloob: It looks like none of the examples care about pr being returned ?

pytradfri/api/aiocoap_api.py
lwis: It will be created when the test request is executed below.

pytradfri/api/aiocoap_api.py
lwis: Observe does.

pytradfri/api/aiocoap_api.py
balloob: Since we know that we will always do a test request, why not create it right away? Every yield, even if it is just to fetch a variable, will suspend the task and wait for the other task to return the result.

lwis: Fair enough, will change later 👍

lwis:
Made that change, turns out aiocoap doesn't let client observations know on `protocol#shutdown`.

The only other thing I'm a little concerned about is that this won't kick in until a manual request is made, as from my testing (which consisted of me unplugging my gateway repeatedly) the observation never drops on it's own - I'm not sure why aiocoap isn't seeing the socket drop? Maybe no heartbeat?

balloob:
It could be no heartbeat. I guess that's a problem we have to fix upstream though?

lwis:
chrysn/aiocoap#79

pytradfri/api/aiocoap_api.py
```diff
@@ -130,6 +152,8 @@ def error_callback(ex):
         ob = pr.observation
         ob.register_callback(success_callback)
         ob.register_errback(error_callback)
+        nonlocal _observations_err_callbacks
```
balloob: I don't think that you have to put a `nonlocal` to access variables in an outer clojure.

lwis: Me neither, but I see `UnboundLocalError: local variable '_protocol' referenced before assignment` if it's not there. How does Python evaluate that?

PR #40, opened 2017-06-27
Fix main

lwis:


balloob:
Can you also update the example commands ?

pytradfri/__main__.py
```diff
@@ -41,19 +42,21 @@ def dump_all():
...
+    print("> light.light_control.set_dimmer(254)*")
+    print("> light.light_control.set_xy_color(254)*")
+    print("> lights[1].light_control.set_dimmer(20)*")
     print("> tasks[0].repeat_days_list")
     print("> groups")
     print("> moods")
     print("> tasks")
     print("> dump_devices()")
     print("> dump_all()")
+    print("""NB: Some commands (marked with *) return a command object
```
balloob: Just wrap the commands in `api(…)` right away

PR #46, opened 2017-08-16
Support for utf8 in lights names

matemaciek:
There may exist more python-ish way to do it. I'm using provided dockerized environment and encountered problem with running this tool when my lights had non-ASCII characters in names..


matemaciek:
It's not a complete fix, I'll try make it such.

PR #48, opened 2017-08-17
Support for setting colour in Kelvin for the whole spectrum

matemaciek:
Inspired by discussion in https://github.com/bwssytems/ha-bridge/issues/570


pytradfri/device.py
```diff
@@ -174,6 +175,9 @@ def set_xy_color(self, color_x, color_y, *, index=0):
             ATTR_LIGHT_COLOR_Y: color_y
         }, index=index)
 
+    def set_kelvin_color(self, kelvins, *, index=0):
+        return self.set_values(dekelvinize(kelvins))
```
lwis: You should include the index here.

matemaciek:
Yes, I've tested it on real IKEA bulbs (-:

pytradfri/kelvin.py
```diff
@@ -0,0 +1,28 @@
...
+  2700: {X: 30140, Y: 26909},
+  4000: {X: 24930, Y: 24694},
+}
+
+
+def can_dekelvinize(kelvins):
+    return min(KNOWN_XY.keys()) <= kelvins <= max(KNOWN_XY.keys())
+
+
+def dekelvinize(kelvins):
```
balloob: Can we rename this? It's not so much a `dekelvinize` as a `kelvin_to_xy` function.

Which means that we should also rename this file to `color.py`, to make it more reusable in the future.

matemaciek: I totally agree.

By the way: I'm still not sure whether we'll need reverse `xy_to_kelvin` function for integration with `home-assistant`.

pytradfri/color.py
```diff
@@ -0,0 +1,48 @@
...
+def can_kelvin_to_xy(k):
+    return k is not None and min(KNOWN_KELVIN) <= k <= max(KNOWN_KELVIN)
+
+
+def can_x_to_kelvin(x):
+    return x is not None and min(KNOWN_X) <= x <= max(KNOWN_X)
+
+
+def kelvin_to_xy(k):
+    if k in KNOWN_XY:
```
lwis: How expensive is this to even bother caching the result?

If it's not negligible, maybe we should consider caching all results in the same map?

matemaciek-9ld: It's not caching - this constant stores known X and Y params for provided color temperatures. Then It uses this constant to estimate X and Y for unknown kelvin values between known values.

lwis: I see, I should have looked at the implementation closer. Sorry.

pytradfri/color.py
```diff
@@ -0,0 +1,48 @@
...
+  4000: {X: 24930, Y: 24694},
+}
+
+KNOWN_KELVIN = KNOWN_XY.keys()
+
+KNOWN_X = [v[X] for v in KNOWN_XY.values()]
+
+
+def can_kelvin_to_xy(k):
+    return k is not None and min(KNOWN_KELVIN) <= k <= max(KNOWN_KELVIN)
```
balloob: Let's remove the `None` check. We should check that before passing it to the util functions.

pytradfri/color.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def can_x_to_kelvin(x):
+    return x is not None and min(KNOWN_X) <= x <= max(KNOWN_X)
+
+
+def kelvin_to_xy(k):
+    if k in KNOWN_XY:
+        return KNOWN_XY[k]
+    else:
```
balloob: No need for the `else` since you return

pytradfri/color.py
```diff
@@ -0,0 +1,48 @@
...
+        higher = KNOWN_XY[higher_k]
+        return {
+            coord: int(offset * higher[coord] + (1 - offset) * lower[coord])
+            for coord in [X, Y]
+        }
+
+
+def x_to_kelvin(x):
+    known = next((k for k, v in KNOWN_XY.items() if v[X] == x), None)
+    if known is not None:
```
balloob: Let's not have `None` checks in color utils.

pytradfri/color.py
```diff
@@ -0,0 +1,48 @@
...
+        return known
+    else:
```
balloob: No else return.

pytradfri/color.py
```diff
@@ -0,0 +1,48 @@
...
+
+KNOWN_X = [v[X] for v in KNOWN_XY.values()]
+
+
+def can_kelvin_to_xy(k):
+    return k is not None and min(KNOWN_KELVIN) <= k <= max(KNOWN_KELVIN)
+
+
+def can_x_to_kelvin(x):
+    return x is not None and min(KNOWN_X) <= x <= max(KNOWN_X)
```
balloob: Why not extract the MIN and MAX and set them to constants instead of calculating them each time?

pytradfri/color.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def kelvin_to_xy(k):
+    if k in KNOWN_XY:
+        return KNOWN_XY[k]
+    else:
+        lower_k = max([kk for kk in KNOWN_KELVIN if kk < k])
```
balloob: You can remove the square brackets around the list comprehension.

pytradfri/color.py
matemaciek: Notice that we check `known`, not `x` and this check is necessary. Of course I can always eliminate this check if I store intermediate list, check its length and return its first element - but it seems to be less clean for me.

balloob: My bad, you are right.

pytradfri/device.py
```diff
@@ -222,6 +226,11 @@ def xy_color(self):
                 self.raw.get(ATTR_LIGHT_COLOR_Y))
 
     @property
+    def kelvin_color(self):
+        if can_x_to_kelvin(self.raw.get(ATTR_LIGHT_COLOR_X)):
```
balloob: Can you include the None check here?

pytradfri/color.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+KNOWN_XY = {
+  2200: {X: 33135, Y: 27211},
+  2700: {X: 30140, Y: 26909},
+  4000: {X: 24930, Y: 24694},
+}
+KNOWN_KELVIN = KNOWN_XY.keys()
+KNOWN_X = [v[X] for v in KNOWN_XY.values()]
+MIN_KELWIN = min(KNOWN_KELVIN)
```
lwis: Should this be `MAX_KELVIN`?

matemaciek:
Yay (-: Is there any release planned soon?

PR #49, opened 2017-08-20
Fix git URL in aiocoap.sh

cosmix:
This patch fixes the tinydtls gitlab link (it was missing the .git suffix).

### DIFF
--- a/script/install-aiocoap.sh
+++ b/script/install-aiocoap.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-git clone --depth 1 https://git.fslab.de/jkonra2m/tinydtls
+git clone --depth 1 https://git.fslab.de/jkonra2m/tinydtls.git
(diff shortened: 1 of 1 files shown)


balloob:
From local testing, both urls work, so it seems the `.git` suffix is optional.

cosmix:
Without the `.git` suffix, I'm getting:

```Cloning into 'tinydtls'...
error: RPC failed; result=22, HTTP code = 404
fatal: The remote end hung up unexpectedly```

on an RPi Model B (512MB) with git version 1.7.10.4. 

Using the `.git` suffix it works as expected.

**Comment by balloob** [comment]
Ah, I tested it with:

```
› git --version
git version 2.11.0 (Apple Git-81)
```

Since both merge, I'll merge this.

balloob:
Thanks.

Issue #50, opened 2017-08-21
Add RGB support


r41d:
(same text as in [general Tradfri discussion](https://github.com/bwssytems/ha-bridge/issues/570#issuecomment-323758488))

Apparently E27 RGB bulbs are now available, today I bought one in an IKEA in Germany.
I didn't find **anything** online, so these gotta be released brand-new.
Here are some photos and screenshots: [http://imgur.com/a/yC5VW](CLICK)
The Android App is able to display the colors that the bulb shows, but it's not able to set them. Setting RGB color only works at the moment by repeatedly pressing the < and > buttons on the physical Tradfri remote.
Will try to do further investigation soon.

EDIT:
- Cycleable colors are: The three normal whites (cold, normal, warm), candlelight, warm amber, dark peach, saturated pink, light purple, light blue. (These color names are shown when you click on New Mood for a category which the RGB bulb is in.)

ggravlingen:
Did you see this? https://github.com/ggravlingen/pytradfri/pull/48

matemaciek:
As soon as I get RGB bulb I'll be able to add support to pytradfri. It's bringing me a lot off fun (-:

schliflo:
@ggravlingen as i understand #48 it is just for the white tones (yellow-ish white to blue-ish white) - the bulbs @r41d mentions are full RGB.

r41d:
@ggravlingen @schliflo correct. I saw the #48 PR earlier, but understood that it was only meant for white spectrum.
Now that IKEA actually released full RGB bulbs, it's the main question how to set arbitrary colors via the gateway :-)
See [my post from April](https://github.com/bwssytems/ha-bridge/issues/570#issuecomment-293905222) where I gathered some theories about the COLOR_X and COLOR_Y parameters, IKEA probably uses the CIE 1931 xy color space, but I think most people would like R, G and B values for their colors, a quick google led me [to this conversion method](http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html) :)

lwis:
@r41d sounds like you're half way to a PR :wink:

ggravlingen:
Great work guys! Closing this issue as it's been resolved through #51..

PR #51, opened 2017-08-23
Add sRGB to xyY conversion

r41d:
This is the foundation needed for #50.
I tested the function with some values and it works great when looking at this image: https://upload.wikimedia.org/wikipedia/commons/3/3b/CIE1931xy_blank.svg
The function uses the sRGB function with D65 white, also I wanted to keep this function minimalistic.
@matemaciek's PR is the base for this, as the RGB color stuff should be integrated in color.py, I think. 

When more sophisticated conversion functions may be needed in the future, https://github.com/gtaylor/python-colormath might be worth a try.
The inverse function to calculate RGB from given xyY would also maybe be useful.

### DIFF
--- a/pytradfri/color.py
+++ b/pytradfri/color.py
@@ -43,8 +43,9 @@ def can_kelvin_to_xy(k):
     return MIN_KELVIN <= k <= MAX_KELVIN
 
 
+# Only used locally to perform normalization of x, y values
 # Scaling to 65535 range and rounding
-def xy2tradfri(x, y):
+def normalize_xy(x, y):
     return (int(x*65535+0.5), int(y*65535+0.5))
 
 
@@ -81,7 +82,7 @@ def kelvin_to_xyY(T, white_spectrum_bulb=False):
     elif T <= 25000:
         y = 3.081758*x**3 - 5.8733867*x**2 + 3.75112997*x - 0.37001483
 
-    x, y = xy2tradfri(x, y)
+    x, y = normalize_xy(x, y)
     return {X: x, Y: y}
 
 
@@ -92,6 +93,30 @@ def xyY_to_kelvin(x, y):
     n = (x/65535-0.3320) / (y/65535-0.1858)
     kelvin = int((-449*n**3 + 3525*n**2 - 6823.3*n + 5520.33) + 0.5)
     return kelvin if MIN_KELVIN <= kelvin <= MAX_KELVIN else None
+
+
+def rgb2xyzA(r, g, b):
+    # Uses CIE standard illuminant A = 2856K
+    # src: http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html
+    # calculation https://gist.github.com/r41d/43e14df2ccaeca56d32796efd6584b48
+    X = 0.76103282*r + 0.29537849*g + 0.04208869*b
+    Y = 0.39240755*r + 0.59075697*g + 0.01683548*b
+    Z = 0.03567341*r + 0.0984595*g + 0.22166709*b
+    return X, Y, Z
+
+
+def rgb2xyzD65(r, g, b):
+    # Uses CIE standard illuminant D65 = 6504K
+    # src: http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html
+    X = 0.4124564 * r + 0.3575761 * g + 0.1804375 * b
+    Y = 0.2126729 * r + 0.7151522 * g + 0.0721750 * b
+    Z = 0.0193339 * r + 0.1191920 * g + 0.9503041 * b
+    return X, Y, Z
+
+
+def xyz2xyY(X, Y, Z):
+    total = X + Y + Z
+    return (0, 0) if total == 0 else normalize_xy(X / total, Y / total)
 
 
 def rgb_to_xyY(r, g, b):
@@ -105,12 +130,5 @@ def rgb_to_xyY(r, g, b):
             return ((val + 0.055) / 1.055) ** 2.4
     r, g, b = map(prepare, (r, g, b))
 
-    # source: http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html
-    CIE_X = 0.4124564*r + 0.3575761*g + 0.1804375*b
-    CIE_Y = 0.2126729*r + 0.7151522*g + 0.0721750*b
-    CIE_Z = 0.0193339*r + 0.1191920*g + 0.9503041*b
-    CIE_sum = CIE_X + CIE_Y + CIE_Z
-
-    (x, y) = (0, 0) if CIE_sum == 0 else \
-        xy2tradfri(CIE_X / CIE_sum, CIE_Y / CIE_sum)
+    x, y = xyz2xyY(*rgb2xyzA(r, g, b))
     return {X: x, Y: y}
--- a/pytradfri/color.py
+++ b/pytradfri/color.py
@@ -10,6 +10,34 @@ MAX_KELVIN = 25000
 MIN_KELVIN_WS = 2200
 MAX_KELVIN_WS = 4000
 
+# Extracted from Tradfri Android App string.xml
+COLOR_NAMES = {
+    '4a418a': 'Blue',
+    '6c83ba': 'Light Blue',
+    '8f2686': 'Saturated Purple',
+    'a9d62b': 'Lime',
+    'c984bb': 'Light Purple',
+    'd6e44b': 'Yellow',
+    'd9337c': 'Saturated Pink',
+    'da5d41': 'Dark Peach',
+    'dc4b31': 'Saturated Red',
+    'dcf0f8': 'Cold sky',
+    'e491af': 'Pink',
+    'e57345': 'Peach',
+    'e78834': 'Warm Amber',
+    'e8bedd': 'Light Pink',
+    'eaf6fb': 'Cool daylight',
+    'ebb63e': 'Candlelight',
+    'efd275': 'Warm glow',
+    'f1e0b5': 'Warm white',
+    'f2eccf': 'Sunrise',
+    'f5faf6': 'Cool white'
+}
+# When setting colors by name via the API,
+# lowercase strings with no spaces are preferred
+COLORS = {name.lower().replace(" ", "_"): hex
+          for hex, name in COLOR_NAMES.items()}
+
 
 def can_kelvin_to_xy(k):
     return MIN_KELVIN <= k <= MAX_KELVIN
--- a/pytradfri/color.py
+++ b/pytradfri/color.py
@@ -1,4 +1,3 @@
-import math
 from .const import (ATTR_LIGHT_COLOR_X as X, ATTR_LIGHT_COLOR_Y as Y)
 
 
@@ -55,17 +54,21 @@ def xy2tradfri(x, y):
 
 
 def kelvin_to_xyY(T):
-    # Sources: "Design of Advanced Color - Temperature Control System for HDTV Applications" [Lee, Cho, Kim]
+    # Sources: "Design of Advanced Color - Temperature Control System
+    #           for HDTV Applications" [Lee, Cho, Kim]
     # and https://en.wikipedia.org/wiki/Planckian_locus#Approximation
     # and http://fcam.garage.maemo.org/apiDocs/_color_8cpp_source.html
     if not (1667 <= T <= 25000):
         return None, None
 
     if T <= 4000:
-        # One number differs on Wikipedia and the paper: 0.2343589 is 0.2343580 on Wikipedia... don't know why
-        x = -0.2661239*(10**9)/T**3 - 0.2343589*(10**6)/T**2 + 0.8776956*(10**3)/T + 0.17991
+        # One number differs on Wikipedia and the paper:
+        #     0.2343589 is 0.2343580 on Wikipedia... don't know why
+        x = -0.2661239*(10**9)/T**3 - 0.2343589*(10**6)/T**2 \
+            + 0.8776956*(10**3)/T + 0.17991
     elif T <= 25000:
-        x = -3.0258469*(10**9)/T**3 + 2.1070379*(10**6)/T**2 + 0.2226347*(10**3)/T + 0.24039
+        x = -3.0258469*(10**9)/T**3 + 2.1070379*(10**6)/T**2 \
+            + 0.2226347*(10**3)/T + 0.24039
 
     if T <= 2222:
         y = -1.1063814*x**3 - 1.3481102*x**2 + 2.18555832*x - 0.20219683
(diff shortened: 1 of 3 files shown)


matemaciek:
Nice one! (-:

Just keep in mind that I have already adjusted home-assistant to support full white spectrum, but it's not yet pushed to PR. If you're using home-assistant, I can push that code faster than I intended.

matemaciek:
BTW, have you tried setting RGB directly as described here: https://github.com/bwssytems/ha-bridge/issues/570#issuecomment-293689166 ? It didn't work for white spectrum bulbs, but maybe it works for RGB bulbs?

r41d:
The `kelvin_to_xyY` function uses the [mathematical computations from "Design of Advanced Color - Temperature Control System for HDTV Applications"](http://www.jkps.or.kr/journal/download_pdf.php?spage=865&volume=41&number=6) for the x and y coordinates, rather than a linear approximation based on some known values from the IKEA App.
Testing this function showed that it lines up with all X and Y values from the KNOWN_XY dictionary, there were some offsets, but they were very small, so this should work for every kelvin value in the range 1667-25000, but I doubt that Trådfri bulbs can actually show all of that ^^
Reverse function still to come...

balloob:
#48 has been merged, please rebase your branch so it only includes the relevant commits.

r41d:
Did the rebasing, I think most stuff here is done, will do some testing when I get home, please don't merge yet.
I didn't want to remove the kelvin-related functions from @matemaciek, the `kelvin_to_xyY` function should, which is based on the math from the paper I found, should cover all those cases. A maintainer may decide to do so.

matemaciek:
I think that `kelvin_to_xyY` should be tested on white spectrum lights - if it's working properly it should replace `kelvin_to_xy`, as it is more precise.

r41d:
**Awesome news**, just got home and tested the `rgb_to_xyY` function, **IIT WORKS** for arbitrary colors, though the colors don't look exactly like I would expect them to, red, green and blue are all kind of "dull" (you would have to see for yourself), but I would say it's enough for a proof-of-concept, algorithms can be tuned later :arrow_right: As soon as the Trådfri Android App has support for RGB bulbs, we might be able to get some useful information from the App.

Unfortunately **no luck** with setting colors [like described here](https://github.com/bwssytems/ha-bridge/issues/570#issuecomment-293689166).
I didn't get anything to work... the only values that get accepted are `f5faf6`, `f1e0b5` or `efd275`, which are the 3 pre-configured white shades.
But at least setting the color via `rgb_to_xyY` works.

As for setting the kelvin via `kelvin_to_xyY`, this works **perfectly** for the whole range, all the way from 1667 to 25000 kelvin. Yes, you can go lower than 2200 and higher than 4000 :D If @matemaciek agrees, I would substitute the kelvin function in `set_kelvin_color` and remove the old `kelvin_to_xy` function, here in this PR.

matemaciek:
@r41d feel free to remove old `kelvin_to_xy`, the less code, the less place for bugs (-; I will test new function  on white spectrum lights, but I guess there is no point going beyond current range on them (they have only two LED colours inside, 2200K and 4000K).

r41d:
@matemaciek: I may have forgotten to mention the visible effect of going below 2200, test it, the color get's much warmer than with lower values. Also for more than 4000, the light get's much colder.
Therefore, I think support for the whole spectrum our formula can should be contained in the library.

matemaciek:
@r41d I assume you're describing behaviour of RGB bulbs - yes, if they're able to show more, we should allow it in API for RGB bulbs. But for white spectrum lights we should limit it, as they're technically unable to show colours outside of this range. That means we need to differentiate between RGB and WS bulbs. For now I only have WS bulbs, but this weekend I'll be visiting Germany, so let's keep our fingers crossed IKEA Dresden have RGB bulbs. Official start of sale is both in Poland and Germany set to 2017-10.

r41d:
@matemaciek: Yes, you're right. I forgot that I was testing with the RGB bulb all the time...
When testing with a 980lm white-spectrum, the function still works, but won't show anything outside of the range of 2200-4000.
But we don't need to differentiate between RGB and WS bulbs: when supplying 1800 for example to a WS bulb, it lights as if I typed 2200, the same for values higher than 4000, the gateway or bulb seem to do boundary checking of the white shade.
Where did you get the official date from?

lwis:
This is looking great, what's the reason for using `xyY` in the func names? 

It would be fantastic if we had support for RGB colours and more precise white spectrum control in the lib and HA before the official app gets support.

matemaciek:
@lwis `xyY` is one of colour spaces: https://en.wikipedia.org/wiki/CIE_1931_color_space#CIE_xy_chromaticity_diagram_and_the_CIE_xyY_color_space

Yeah, it would be very nice to have it before official app (-: I've already added white spectrum support to my HA, but I'm waiting with PR until newer `pytadfri` is released: https://github.com/home-assistant/home-assistant/compare/dev...matemaciek:dev

r41d:
I found a new way of setting some predefined colors directly without any hazzle or calculation.
When looking at the Tradfri App at `res/values-en-rUS/strings.xml` I found.
```xml
    <string name="_4a418a">Blue</string>
    <string name="_6c83ba">Light Blue</string>
    <string name="_8f2686">Saturated Purple</string>
    <string name="_a9d62b">Lime</string>
    <string name="_c984bb">Light Purple</string>
    <string name="_d6e44b">Yellow</string>
    <string name="_d9337c">Saturated Pink</string>
    <string name="_da5d41">Dark Peach</string>
    <string name="_dc4b31">Saturated Red</string>
    <string name="_dcf0f8">Cold sky</string>
    <string name="_e491af">Pink</string>
    <string name="_e57345">Peach</string>
    <string name="_e78834">Warm Amber</string>
    <string name="_e8bedd">Light Pink</string>
    <string name="_eaf6fb">Cool daylight</string>
    <string name="_ebb63e">Candlelight</string>
    <string name="_efd275">Warm glow</string>
    <string name="_f1e0b5">Warm white</string>
    <string name="_f2eccf">Sunrise</string>
    <string name="_f5faf6">Cool white</string>
```
Searching for those number with underscore led me to `com.ikea.tradfri.lighting.p054i.p089a.C1384a` (I decompiled the App with JADX) which has this part:
```java
        this.f3891b = new HashMap();
        this.f3891b.put("f5faf6", new C1386c(0.3804d, 0.3804d, "f5faf6", 0.54d));
        this.f3891b.put("f1e0b5", new C1386c(0.4599d, 0.4106d, "CCT_LIGHT_NEUTRAL", 0.61d));
        this.f3891b.put("efd275", new C1386c(0.5056d, 0.4152d, "efd275", 0.66d));
        this.f3891b.put("dcf0f8", new C1386c(0.3221d, 0.3317d, "dcf0f8", 0.45d));
        this.f3891b.put("eaf6fb", new C1386c(0.3451d, 0.3451d, "eaf6fb", 0.48d));
        this.f3891b.put("f5faf6", new C1386c(0.3804d, 0.3804d, "f5faf6", 0.54d));
        this.f3891b.put("f2eccf", new C1386c(0.4369d, 0.4041d, "f2eccf", 0.59d));
        this.f3891b.put("CCT_LIGHT_NEUTRAL", new C1386c(0.4599d, 0.4106d, "CCT_LIGHT_NEUTRAL", 0.61d));
        this.f3891b.put("efd275", new C1386c(0.5056d, 0.4152d, "efd275", 0.66d));
        this.f3891b.put("ebb63e", new C1386c(0.5516d, 0.4075d, "ebb63e", 0.68d));
        this.f3891b.put("e78834", new C1386c(0.58d, 0.38d, "e78834", 0.69d));
        this.f3891b.put("e57345", new C1386c(0.58d, 0.35d, "e57345", 0.67d));
        this.f3891b.put("da5d41", new C1386c(0.62d, 0.34d, "da5d41", 0.7d));
        this.f3891b.put("dc4b31", new C1386c(0.66d, 0.32d, "dc4b31", 0.73d));
        this.f3891b.put("e491af", new C1386c(0.5d, 0.28d, "e491af", 0.57d));
        this.f3891b.put("e8bedd", new C1386c(0.45d, 0.28d, "e8bedd", 0.53d));
        this.f3891b.put("d9337c", new C1386c(0.5d, 0.24d, "d9337c", 0.55d));
        this.f3891b.put("c984bb", new C1386c(0.34d, 0.19d, "c984bb", 0.38d));
        this.f3891b.put("8f2686", new C1386c(0.31d, 0.12d, "8f2686", 0.33d));
        this.f3891b.put("4a418a", new C1386c(0.17d, 0.05d, "4a418a", 0.18d));
        this.f3891b.put("6c83ba", new C1386c(0.2d, 0.1d, "6c83ba", 0.22d));
        this.f3891b.put("a9d62b", new C1386c(0.4099999964237213d, 0.5099999904632568d, "a9d62b", 0.654d));
        this.f3891b.put("d6e44b", new C1386c(0.44999998807907104d, 0.4699999988079071d, "d6e44b", 0.65d));
```
The `C1386c` class seems to just 3 doubles and 1 string, which somehow represents a color.

But the classes are not really needed, they are just here for reference, the cool part is that you can just take one of the hex values and shove them into `set_hex_color` and it works.
I included the extracted color names and values in color.py and wrote a function to set predefined colors, will push soon.

matemaciek:
Those numbers may be scaled x and y, we should verify it (-: We should also investigate rgb bulb behaviour - how it reacts to rgb and xy values between known values.

pytradfri/color.py
```diff
@@ -1,48 +1,73 @@
...
+def xy2tradfri(x, y):
+    return (int(x*65535+0.5), int(y*65535+0.5))
+
+
+def kelvin_to_xyY(T):
+    # Sources: "Design of Advanced Color - Temperature Control System
+    #           for HDTV Applications" [Lee, Cho, Kim]
+    # and https://en.wikipedia.org/wiki/Planckian_locus#Approximation
+    # and http://fcam.garage.maemo.org/apiDocs/_color_8cpp_source.html
+    if not (1667 <= T <= 25000):
```
balloob: We should raise when invalid data gets passed in. Because otherwise we end up failing anyway because we get `None`.

```python
    if not (MIN_KELVIN <= T <= MAX_KELVIN):
        raise ValueError('Kelvin needs to be between {} and {}'.format(
            MIN_KELVIN, MAX_KELVIN))
```

pytradfri/color.py
```diff
@@ -1,48 +1,73 @@
...
+
+    if T <= 2222:
+        y = -1.1063814*x**3 - 1.3481102*x**2 + 2.18555832*x - 0.20219683
+    elif T <= 4000:
+        y = -0.9549476*x**3 - 1.37418593*x**2 + 2.09137015*x - 0.16748867
+    elif T <= 25000:
+        y = 3.081758*x**3 - 5.8733867*x**2 + 3.75112997*x - 0.37001483
+
+    x, y = xy2tradfri(x, y)
+    return {X: x, Y: y}
```
balloob: The methods are not consistent with their return types. Some methods return dictionaries, others tuples. We should pick one and stick with it.

pytradfri/color.py
```diff
@@ -1,48 +1,73 @@
...
+
+
+def xyY_to_kelvin(x, y):
+    # This is an approximation, for information, see the source.
+    # Source: https://en.wikipedia.org/wiki/Color_temperature#Approximation
+    # Input range for x and y is 0-65535
+    n = (x/65535-0.3320) / (y/65535-0.1858)
+    kelvin = int((-449*n**3 + 3525*n**2 - 6823.3*n + 5520.33) + 0.5)
+    return kelvin if MIN_KELVIN <= kelvin <= MAX_KELVIN else None
```
balloob: Why would we check for MIN and MAX KELVIN here ? it's that Kelvin right?

balloob:
This is awesome work! I hit approve a bit too fast, actually had some comments.

I will do a new release of pytradfri when this PR is merged @matemaciek, so support can land in Home Assistant 0.53

pytradfri/color.py
r41d: Can do. Will put this in

pytradfri/color.py
r41d: Your're right. This get's fixed by raising when Kelvin is outside range

pytradfri/color.py
r41d: This reverse function for xyY to Kelvin is an approximation (https://en.wikipedia.org/wiki/Color_temperature#Approximation) which is not 100% accurate.
I just didn't want to return kelvin values for which the kelvin to xyY function wouldn't work (because that one raises when a values outside of the range is passed), this is why I only wanted to return the kelvins when they are in this range.

pytradfri/color.py
balloob: I meant that sometimes you return a dictionary, other times you return a tuple

pytradfri/color.py
balloob: It's not up to this function to decide if it's a valid Kelvin, this function should only offer an approximation. The device class should check it with `can_kelvin_to_xy` to decide to return `None` or not.

pytradfri/color.py
r41d: `kelvin_to_xyY` always returns a dict.
Did you mean `xy2tradfri`? This function is only used locally to encapsulate some calculations.

pytradfri/color.py
r41d: This is the reverse function, so `can_kelvin_to_xy` is not applicable, the function that would be applicable is `can_xy_to_kelvin`, which doesn't exist at the moment because I thought that by returning None if the conversion shouldn't succeed I could conserve this function.
Should I create `can_xy_to_kelvin`?

pytradfri/color.py
r41d: Where is a tuple returned?

r41d:
The latest commit outsources the rgb->xyz calculation into a function, actually there's two of them, one for CIE standard illuminant D65 and for A.
I calculated the matrix for A myself: https://gist.github.com/r41d/43e14df2ccaeca56d32796efd6584b48
The idea behind using CIE standard illuminant A was that 2700K is the "normal" white for Tradfri bulbs, so I think this should be the standard white for rgb->xyz conversion here.
I didn't find a formula to convert a Kelvin value to XYZ chromaticity coordinates, so I used the standard white A, which is 2856K because it's closest to it. (https://en.wikipedia.org/wiki/Standard_illuminant#White_points_of_standard_illuminants)
Standard white A gives me (subjectively) better looking results than with the old D65-based matrix.

Maybe someone with more knowledge about CIE, XYZ, RGB and such might improve my code.

I think this PR is pretty much done. Waiting for final feedback.

pytradfri/color.py
balloob: No, you can use `can_kelvin_to_xy` to avoid returning Kelvin inside the Ikea device. However, for color utility functions, we should not be concerned with what range IKEA supports, but instead just do our best to give an approximation.

r41d: Included what I think you meant in commit 768c8eb :)

ggravlingen:
Great work, look forward to buying a RGB-bulb and start using this!

ggravlingen:
@r41d and @matemaciek I just made a release to pypi based on your contributions to the library. 🎉  Feel free to integrate into to the home automation platform of your choice :)

PR #52, opened 2017-08-30
Version bump

matemaciek:
Shouldn't we update version here?


ggravlingen:
Yes, I changed it when I released to pypi but forgot to commit.

PR #60, opened 2017-09-30
Fix upstream changes, pin commit for aiocoap

lwis:


pytradfri/api/aiocoap_api.py
```diff
@@ -27,10 +27,10 @@ def _get_psk(self, host, port):
 
 
 def _patched_datagram_received(self, data, addr):
-    self._dtls_socket.handleMessage(self._connection, data, 0)
+    self.parent._dtls_socket.handleMessage(self.parent._connection, data, 0)
```
balloob: Should we see to get this fixed upstream?

balloob:
I'm on the move right now so can't do a release.

ggravlingen:
I released the change and pushed 2.2.2 to pypi @balloob @lwis

lwis:
@ggravlingen updated in HA :+1:

PR #61, opened 2017-10-01
Ensure consistent API return type (breaking)

lwis:


pytradfri/api/aiocoap_api.py
```diff
@@ -119,10 +120,10 @@ def _execute(api_command):
         return api_command.result
 
     @asyncio.coroutine
-    def request(*api_commands):
+    def request(api_commands):
         """Make a request."""
-        if len(api_commands) == 1:
-            result = yield from _execute(api_commands[0])
+        if not isinstance(api_commands, collections.Iterable):
```
balloob: Please add a test for this. I would explicitly check for lists, because generators can be used for async in python 3.4

lwis:
I'm sure I'm missing something writing those tests, but I couldn't find a way to mock an inner function (_execute). I should have switched the factories to classes when I had the chance.

balloob:
🐬  Ok to merge when you get the tests to pass 👍  stoked this all works without running custom scripts anymore!

lwis:
Done, build is passing now. aiocoap requires 3.4.4 minimum, so I had to bump the 3.4 version.

I experimented (as can be seen in the commits) with using the [tinydtls] extra on aiocoap, but when I ran `pip install .[async]` it complained about not being able to find the tinydtls extra on the egg. So I just declared a dependency on DTLSSocket directly from here, once a new verison of aiocoap is published this should all be much cleaner as we don't need a dep link.

ggravlingen:
Yaay! @lwis please feel free to release 3.0! :1st_place_medal:

lwis:
Woo! Not at my machine right now, but will release if nobody beats me to
it! Will hopefully get around to updating HA this weekend.

ggravlingen:
Sorry, got a bit carried away. :blush:

I'm happy to push to pypi if you need assistance with that admin stuff after you've done the release.

Issue #54, opened 2017-09-03
Issues with necessary gateway restart


melsom:
Hello,

As many others are experiencing. When you reach a certain number of bulbs, the gateway needs restarting every 1-2 days. I use IKEA Trådfri with Home Assistant, and feature wise it works great.

Thing is, whenever I reboot the gateway to be able to reach the bulbs, all light groups are switched to off in Home Assistant. Even though they are on. This happens every time.

Home Assistant says it supports local push, which means updates should be pushed automatically. Any way to fix this? Waiting for IKEA to fix their gateway could take some time.. This just started happening when reaching bulb number 20.

lwis:
I've always thought the gateway firmware has a memory leak somewhere. Particularly when you're making requests every 15 seconds, this is quite noticeable.

As for the groups, HA will query the gateway, if the gateway reports the group is off, that's all the information we have.

However, if the app (after a restart) is reporting them as on, there may be a bug in the library.

With the switch to observations imminent, I'm hoping this will be more reliable.

melsom:
When checking the TRÅDFRI app after reboot, it shows the groups as being on, and those off obviously off. But HA does not seem to be able to pull this information. 

So there might be a bug somewhere. Please let me know if there is anything I can do to help.

hanpal:
One strange thing is that reboot (power off/on) worked fine for me for a long time, fully automated. But since the last month it doesn't, only factory reset works. I've automated this also but this required soldering wires to the circuit board. My automation always tries power off/on and then waits a while and checks if it works, if not, factory reset. Factory reset is initiated almost every day. It may be connected to some pytradfri problem as well but each time that connection is lost, it is also lost using the app so it seems that the gateway has been affected at least.

Memory leaks are highly suspected, probably both in RAM and perhaps also some problems with logs written to flash memory, missing or faulty round robin handling.

lwis:
Have you tried not running pytradfri for a few days to see if the issue persists?

hanpal:
Well, not running pytradfri would most certainly remove the issue but also completely remove the functionality I want. Not interested in using only a controller (remote) or the app, since I have a fully automated lighting system using Tradfri.

spektren:
reboot and factory reset can be issued by network. 

REBOOT: 
coap-client -m post -u "Client_identity" -k "**key**" "coaps://192.168.178.71:5684/15011/9030"

!!!RESET!!! (everything like pairing, groups, etc. will be gone): 
[...]

PR #66, opened 2017-10-13
added GW functions reboot() and factory_defaults()

spektren:
Might be useful to some users (see #54). 

I'm not sure about the return behaviour if the command fails. 

After a reboot is issued the enumeration for the devices in pytradfri might change. Also the GW doesn't seem to transport any commands for quite some time (5mins?).

### DIFF
--- a/pytradfri/gateway.py
+++ b/pytradfri/gateway.py
@@ -155,7 +155,7 @@ class Gateway:
 
     def reboot(self):
         """
-        reboot GW
+        Reboot the Gateway
 
         Returns a Command.
         """
@@ -163,10 +163,10 @@ class Gateway:
         return Command('post',
                        [ROOT_GATEWAY, ATTR_GATEWAY_REBOOT])
 
-    def factory_defaults(self):
+    def factory_reset(self):
         """
-        resets GW to factory default
-        warning: all data in GW is lost (pairing, groups, ...)
+        Resets the Gateway to factory defaults.
+        WARNING: All data in Gateway is lost (pairing, groups, etc)
 
         Returns a Command.
         """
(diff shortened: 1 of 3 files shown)


pytradfri/gateway.py
```diff
@@ -150,6 +151,36 @@ def process_result(result):
         return Command('get', [ROOT_SMART_TASKS, task_id],
                        process_result=process_result)
 
+    def reboot(self):
+        """
+        Reboot gateway.
+
+        Returns a Command.
+        """
+        def process_result(result):
```
lwis: Do you need to wrap the result?

spektren: I didnt' succeed in getting any response from the GW. Not idea how to provoke a response (namely negative response). That's why I put the wrapper in there but left it empty in the end. 

Should I remove them?

lwis: Yeah, it doesn't add anything.

pytradfri/gateway.py
```diff
@@ -150,6 +151,26 @@ def process_result(result):
...
                        process_result=process_result)
 
+    def reboot(self):
+        """
+        Reboot gateway.
+
+        Returns a Command.
+        """
+        return Command('post',
+                       PATH_GATEWAY_INFO[0] + '/' + ATTR_GATEWAY_REBOOT)
```
lwis: Wondering if we should refactor PATH_GATEWAY_INFO in const.py, and have reboot and wipe be constants with paths there instead? Doing `[0]` isn't very clean.

ggravlingen:
@spektren Is it a wipe of settings or factory reset?

spektren:
Aren't both terms resulting in the same state of virginity (unconfigured but ready to use)? 
As far as I can tell the gw was like when I unboxed it for very first time...

I wanted to avoid 'reset' to indicate clearly that all configurations will be lost when 'wipe' is executed. 

If preferred, I'll change the wording in the code.

ggravlingen:
@spektren In that case I'd prefer factory_reset() as the connotation is very clear: don't expect anything of the gateway's current state to remain.

spektren:
@lwis @ggravlingen, thanks for your inputs. 

Please have a look at the changes.

lwis:
Thanks @spektren :crocodile:

ggravlingen:
Just released 2.2.3 including this PR.

ggravlingen:
2.2.3 has also been pushed to pypi. Sorry about the delay, had some computer issues.

Issue #65, opened 2017-10-08
Move discovered tradfri-behaviour specific code from home-assistant to pytradfri


matemaciek:
Suggested specification:
- `LightControl` will receive new methods:
  - `can_set_color` - returns whether controlled light supports arbitrary color. If so, all of following should change color of light as requested: `set_hex_color`, `set_xy_color`, and `set_rgb_color`. The range of supported colors is not specified, but if some color works in one space and can be specified in other space it should also work there.
  - `can_set_kelvin` - returns whether controlled light supports color temperature. If so, `set_kelvin_color` should change color of light as requested. The range of supported tempertures is defined by methods `min_kelvin` and `max_kelvin`.
  - `min_kelvin` - explained above.
  - `max_kelvin` - explained above.
- `Light` will satisfy following speciffication:
  -  `kelvin_color`, `xy_color`, `hex_color` and `rgb_color` will return colors coresponding to actual state of the bulb, regardles of bulb type (white / white spectrum / color & white spectrum). Exception: `kelvin_color` on colors outside of white spectrum may return `None`. Boundaries of white spectrum in colored spaces are not strictly defined.

matemaciek:
I'm currently working on described changes, so anyone with enough power can assign this issue to me.

ggravlingen:
Great initiative! 🎉

balloob:
I like it!

matemaciek:
Just imagine how neat `home-assistant`'s code responsible for tradfri will get (-:

ggravlingen:
I guess #67 is a pre-requisite for solving this issue?

matemaciek:
#67 is a first part - whell, technically second - taking care of `Light` behaviour.

ggravlingen:
@matemaciek are we ready to close this issue now?

matemaciek:
@ggravlingen I think we need to sort up setters also - since some of them are doing conversions, and some are just setting API parameters. Especially `set_hex_color` is only working for some hex values, although we now how to set them for more values.

ggravlingen:
Ah, I see, that makes sense. Would you be interested in also adding these setters to the lib since you know the structure behind the getters?

matemaciek:
Yes, I'm planning to do so (-: What do you think about suffixing "magic" setters with `_inferred`, as we did in getters?

matemaciek:
BTW I think that setting hex/rgb should also set brightness, not only x and y. @r41d , what do you think?

[...]

PR #67, opened 2017-10-14
Complying to #65 - reading parameters of different bulbs types

matemaciek:


pytradfri/device.py
```diff
@@ -230,12 +230,27 @@ def dimmer(self):
 
     @property
     def hex_color(self):
-        return self.raw.get(ATTR_LIGHT_COLOR)
+        raw_color = self.raw.get(ATTR_LIGHT_COLOR)
+        if raw_color is not None and len(raw_color) == 6:
+            return raw_color
+        (x, y) = self.xy_color
+
+        def scale(val):
```
balloob: you're defining a function that you're not using?

pytradfri/device.py
```diff
@@ -230,12 +230,27 @@ def dimmer(self):
 
     @property
     def hex_color(self):
-        return self.raw.get(ATTR_LIGHT_COLOR)
+        raw_color = self.raw.get(ATTR_LIGHT_COLOR)
+        if raw_color is not None and len(raw_color) == 6:
+            return raw_color
+        (x, y) = self.xy_color
```
balloob: I'm not sure if I'm 100% happy with this.

This property used to be a direct 1:1 mapping with the underlying data from the API. I think that it should stay like that. If we want to offer smarter things on top, maybe we can offer it as a utility function?

matemaciek: It's true for all `hex_color`, `xy_color` and `kelvin_color`. Maybe each of them need to have two flavours - "raw" and "best guess"? Maybe boolean parameter? Than it can default to one of flavours - which one you'd think would be better?

matemaciek: Hm, they're properties, so it conflicts with parametrizing them. So two sets of properties maybe? `[hex|xy|kelvin]_color_[raw|guess]`? Or `[hex|xy|kelvin]_color[_best_guess]`, or even `[hex|xy|kelvin]_color[_raw]`. I also don't have a good suffix for this new property - `guess`, `best_guess`, `estimate`? Neither captures exact behaviour, which is "If I have a value from API, return it, else estimate from other values from API".

ggravlingen: Are you not talking about something along the line of an inferred value?
https://en.wikipedia.org/wiki/Inference

matemaciek: Yes, inference is the word, thanks! So how do you find a set of properties: `[hex|xy|kelvin]_color_[raw|inferred]`?

ggravlingen: Np! :)

I recon @balloob or @lwis has a more clear view on naming conventions than I do.

matemaciek: OK, let's try to put that all together...

ggravlingen: If you want to continue working on your code now I suggest going for ```[hex|xy|kelvin]_color_[raw|inferred]```. It's not a biggie to update if they suggest other names. :)

matemaciek:
I went with `[hex|xy]_color[_raw]` and `kelvin_color`. Advantages:
- if user knows that he wants raw value, he explicitely asks for it.
- if user just want the value, we're doing the best we can; even if lightbulb has one fixed color, this color has some temperature, although it doesn't inform about it via API
- we're backward compatible

pytradfri/device.py
balloob: let's not call the raw one raw. It's just xy_color, hex_color etc. Then we don't need a breaking change either!

matemaciek: But then we have inconsistency in `kelvin_color`, which is definitely inferred. And this change would be breaking because homeassistant checks bulb features by verifing if `hex_color` makes sense? So maybe it would be good to introduce breaking change, together with `can_set_`, to introduce more order? I'll then add `can_set_` to this PR soon.

balloob: Breaking with Home Assistant is fine as HASS uses pinned requirements. If we messed up with `kelvin_color`, let's fix it.

matemaciek: OK, so which convention should we follow:
1) `[hex|xy]_color[_raw]` and `kelvin_color`
2) `[hex|xy]_color[_inferred]` and `kelvin_color_inferred`
3) `[hex|xy]_color_[raw|inferred]` and `kelvin_color_inferred`
?

balloob: I don't follow your convention list. This is what I would prefer:

- I want, by default, property names to reflect what is returned from the API. xy_color comes directly from the returned dictionary. If we don't do this, the user can no longer trust if some data is there or that we're trying to come up with an answer, which might not fit their use case
- I want the property name to reflect if we're doing magic to make it work. So add `_inferred`.

matemaciek: OK, so what's you're describing is I guess 2) `[hex|xy]_color[_inferred]` and `kelvin_color_inferred`. which means 5 properties:
- `hex_color`,
- `xy_color`,
- `hex_color_inferred`,
- `xy_color_inferred`,
- `kelvin_color_inferred`.
I'll change properties name that way.

ggravlingen:
Guess we're good to merge here? @balloob

ggravlingen:
Merged and pushed to pypi as 3.0.1. Thanks @matemaciek for your contribution!

PR #68, opened 2017-10-15
Update device.py

rudyvan:
If the device is an input instead of a lamp then self.raw is None and has_light_control should return False.
This change caters for that scenario.

### DIFF
--- a/pytradfri/device.py
+++ b/pytradfri/device.py
@@ -48,7 +48,8 @@ class Device(ApiResource):
 
     @property
     def has_light_control(self):
-        return (self.raw is not None and len(self.raw.get(ATTR_LIGHT_CONTROL,"")) > 0)
+        return (self.raw is not None and 
+                len(self.raw.get(ATTR_LIGHT_CONTROL,"")) > 0)
 
     @property
     def light_control(self):
--- a/pytradfri/device.py
+++ b/pytradfri/device.py
@@ -48,7 +48,7 @@ class Device(ApiResource):
 
     @property
     def has_light_control(self):
-        return (self.raw is not None and 
+        return (self.raw is not None and
                 len(self.raw.get(ATTR_LIGHT_CONTROL, "")) > 0)
 
     @property
--- a/pytradfri/device.py
+++ b/pytradfri/device.py
@@ -49,7 +49,7 @@ class Device(ApiResource):
     @property
     def has_light_control(self):
         return (self.raw is not None and 
-                len(self.raw.get(ATTR_LIGHT_CONTROL,"")) > 0)
+                len(self.raw.get(ATTR_LIGHT_CONTROL, "")) > 0)
 
     @property
     def light_control(self):
--- a/pytradfri/device.py
+++ b/pytradfri/device.py
@@ -48,8 +48,7 @@ class Device(ApiResource):
 
     @property
     def has_light_control(self):
-        return (ATTR_LIGHT_CONTROL in self.raw and
-                len(self.raw.get(ATTR_LIGHT_CONTROL)) > 0)
+        return (self.raw is not None and len(self.raw.get(ATTR_LIGHT_CONTROL,"")) > 0)
 
     @property
     def light_control(self):


ggravlingen:
Yay! :+1: Let's try to make this thing past the check.

pytradfri/device.py
```diff
@@ -48,8 +48,7 @@ def reachable(self):
 
     @property
     def has_light_control(self):
-        return (ATTR_LIGHT_CONTROL in self.raw and
-                len(self.raw.get(ATTR_LIGHT_CONTROL)) > 0)
+        return (self.raw is not None and len(self.raw.get(ATTR_LIGHT_CONTROL,"")) > 0)
```
ggravlingen: Travis returned these errors:

```The command "py.test" exited with 0.
1.60s$ flake8
./pytradfri/device.py:51:77: E231 missing whitespace after ','
./pytradfri/device.py:51:80: E501 line too long (86 > 79 charact```

Comment (ggravlingen) [reply pytradfri/device.py]: Which is interpreted as you need to break the line into two as a row many not exceed 79 chars and you need to write ```(ATTR_LIGHT_CONTROL, "")```

Comment (ggravlingen) [reply pytradfri/device.py]: When I started doing this, I quickly realized the power of installing pylint and then running:
```pylint device.py```

It will list the syntax errors (if any) in your code. Travis runs the same checks and will not green flag the change until all tests have pasts. Green flags is, in turn, needed before a merge of the PR is done.

File: pytradfri/device.py [hunk 2]
```diff
@@ -48,7 +48,8 @@ def reachable(self):
 
     @property
     def has_light_control(self):
-        return (self.raw is not None and len(self.raw.get(ATTR_LIGHT_CONTROL,"")) > 0)
+        return (self.raw is not None and
```
ggravlingen: Trailing whitespace after "and"

pytradfri/device.py
```diff
@@ -48,7 +48,8 @@ def reachable(self):
 
     @property
     def has_light_control(self):
-        return (self.raw is not None and len(self.raw.get(ATTR_LIGHT_CONTROL,"")) > 0)
+        return (self.raw is not None and 
+                len(self.raw.get(ATTR_LIGHT_CONTROL,"")) > 0)
```
ggravlingen: no whitespace after comma

ggravlingen:
One down, two to go :+1: If you click "Details" where it says "The Travis CI build failed" and then click one of the rows, for instance where it reads Python 3.6, and then scroll down to where it says ```The command "py.test" exited with 0.``` you can see why the checks fail.

ggravlingen:
Great! We should also add a test for the case where the device is an input instead of a lamp just to make sure the code works as expected. Then we're good to merge!

ggravlingen:
@rudyvan I'm happy to help out if you want some aid in adding the test

rudyvan:
I would think that a sensor does not have the lightcontrol attribute in the raw variable, so the updated test already caters for that scenario.    I would not know what more differentiate between an input and a light and would make the test more complete.   I read somewhere that the tradfri system does not return input data back to the gateway.

ggravlingen:
Yes, you are probably right, just want to make sure I'm not messing anything up here. :smile:  If I understand correctly, your ROOT_DEVICES (15001) looked something like this?

```
{  
   "9054":0,
   "9001":"TRADFRI remote control",
   "5750":0,
   "9002":1490985070,
   "9020":1505761229,
   "9003":65536,
   "9019":0,
   "3":{  
      "0":"IKEA of Sweden",
      "6":3,
      "1":"TRADFRI remote control",
      "2":"",
      "3":"1.2.214",
      "9":60
   },
   "15009":[  
      {  
         "9003":0
      }
   ]
}
```

rudyvan:
my devs look like the following:
```text
[<65536 - motion_sensor (TRADFRI motion sensor)>,
 <65538 - guest_light (TRADFRI bulb E27 W opal 1000lm)>, 
<65537 - remote_control (TRADFRI remote control)>]
```
how do i get my ROOT_DEVICES?

ggravlingen:
This actually serves as a pretty good example (the initial analysis that inspired this lib):
https://bitsex.net/software/2017/coap-endpoints-on-ikea-tradfri/

So, what you listed is the content of ROOT_DEVICES  (15001):

```[<65536 - motion_sensor (TRADFRI motion sensor)>,
 <65538 - guest_light (TRADFRI bulb E27 W opal 1000lm)>, 
<65537 - remote_control (TRADFRI remote control)>]```

I'm thinking a test to make sure your code works as expected might then look something like this, where I'm uncertain what has_light_control returns?


```
SANS_LIGHT =
{  
   "9054":0,
   "9001":"TRADFRI remote control",
   "5750":0,
   "9002":1490985070,
   "9020":1505761229,
   "9003":65536,
   "9019":0,
   "3":{  
      "0":"IKEA of Sweden",
      "6":3,
      "1":"TRADFRI remote control",
      "2":"",
      "3":"1.2.214",
      "9":60
   },
   "15009":[  
      {  
         "9003":0
      }
   ]
}


def test_device_remote_only():
    dev = Device(SANS_LIGHT)

    assert dev.has_light_control == xxx

```

rudyvan:
Dear Patrick,

what i understand you mean is to check if there is ONLY ONE endpoint in the gateway at 15001 which is either a remote or a motion sensor but no light.

Indeed the has_light_control() code should work in these circumstances on that single non bulb device. 
I think the way it is written should already work as it tests the individual device, a bulb or something else.

Adding an assert statement is not needed in my opinion.

KR Rudy

ggravlingen:
You are right, my apologies! Your improvement is now released on pypi as 3.0.2. Thank you for contributing to pytradfri! :+1:

Issue #76, opened 2017-10-26
Travis returns error in device.py


ggravlingen:
```
The command "py.test" exited with 0.
1.06s$ flake8
./pytradfri/device.py:219:9: E722 do not use bare except'
The command "flake8" exited with 1.
cache.2
store build cache
```

ggravlingen:
@r41d this error started showing up now, but wasn't returned by Travis when your PR was first merged. Can you please elaborate what cases you throw into the except-state?

Thanks!

```
    def set_predefined_color(self, colorname, *, index=0):
        try:
            color = COLORS[colorname.lower().replace(" ", "_")]
        except:
            pass
        else:
            return self.set_hex_color(color, index=index)
```

PR #78, opened 2017-10-28
Fixing device.py so that Travis doesn't fail, added test and support for coverall

ggravlingen:
In the current prod version of the code, Travis throws an error in device.py (#76). Oddly, the error wasn't there when the affected lines of code was first merged. This is a fix for that error so that Travis can build again.

I've also added a test for the set_predefined_color-method and support for coverall.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,3 +22,5 @@ language: python
 script:
   - py.test
   - flake8
+after_success:
+  - coveralls
--- a/pytradfri/device.py
+++ b/pytradfri/device.py
@@ -217,7 +217,7 @@ class LightControl:
         try:
             color = COLORS[colorname.lower().replace(" ", "_")]
         except:
-            pass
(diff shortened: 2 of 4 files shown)


ggravlingen:
@balloob or @lwis can you have a look at this solution to the problem please?

lwis:
My preference would be to maintain the existing behaviour, catch the KeyError and do nothing, but log a warning that the colour could not be resolved.

lwis:
You need to update your test, and I think the block should look more like;

```
 def set_predefined_color(self, colorname, *, index=0):
        try:
        except KeyError:
            _LOGGER.debug('Could not match color name')
```

ggravlingen:
@lwis Thanks! Took me a while to figure out how to update the test but this looks like it's passing through the gates of Travis.

balloob:
Please use clean branches. Seems like it contains a lot of unrelated commits

PR #80, opened 2017-10-28
Fixing device.py so that Travis doesn't fail and added test

ggravlingen:
In the current prod version of the code, Travis throws an error in device.py (#76). Oddly, the error wasn't there when the affected lines of code was first merged. This is a fix for that error so that Travis can build again. I've also added a test for the set_predefined_color-method.

### DIFF
--- a/pytradfri/device.py
+++ b/pytradfri/device.py
@@ -20,6 +20,9 @@ from .color import can_kelvin_to_xy, kelvin_to_xyY, xyY_to_kelvin, rgb_to_xyY,\
     xy_brightness_to_rgb, COLORS, MIN_KELVIN, MAX_KELVIN,\
     MIN_KELVIN_WS, MAX_KELVIN_WS
 from .resource import ApiResource
+import logging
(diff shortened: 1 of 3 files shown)


ggravlingen:
@balloob created a new, clean, branch with minimum commits

pytradfri/device.py
```diff
@@ -216,10 +219,9 @@ def set_kelvin_color(self, kelvins, *, index=0):
     def set_predefined_color(self, colorname, *, index=0):
         try:
             color = COLORS[colorname.lower().replace(" ", "_")]
-        except:
-            pass
-        else:
             return self.set_hex_color(color, index=index)
+        except KeyError:
```
balloob: I don't like swallowing errors. Because now as a caller, I won't know if my call succeeded and that means that I can't communicate this to my users. I think that we should never catch but instead raise as much as possible. That way we leave it to developers that use this lib if they want to swallow errors or show them to users.

PR #88, opened 2017-10-29
Raise error instead of writing to log

ggravlingen:
Coverage dropped because I've added the pytradfri directory rather than only testing the tests-folder.

### DIFF
--- a/pytradfri/device.py
+++ b/pytradfri/device.py
@@ -221,7 +221,7 @@ class LightControl:
             return self.set_hex_color(color, index=index)
         except KeyError:
             raise ColorError('Unable to match %s in COLOR_NAMES',
-                colorname)
+                             colorname)
 
     def set_rgb_color(self, r, g, b, *, index=0):
         return self.set_values(rgb_to_xyY(r, g, b), index=index)
--- a/pytradfri/device.py
+++ b/pytradfri/device.py
@@ -220,7 +220,7 @@ class LightControl:
             color = COLORS[colorname.lower().replace(" ", "_")]
             return self.set_hex_color(color, index=index)
         except KeyError:
-            PyTradFriError("Could not match color name in COLOR_NAMES")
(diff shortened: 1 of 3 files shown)


tests/test_light.py
```diff
@@ -233,7 +233,3 @@ def test_setters():
     cmd = Device(LIGHT_CWS).light_control \
         .set_predefined_color('Warm glow')
     assert cmd.data == {'3311': [{'5706': 'efd275'}]}
-
-    cmd = Device(LIGHT_CWS).light_control \
```
balloob: Add a test that tests that it raises:

```
import pytest
from pytradfri import error

with pytest.raises(error.InvalidColor):
    Device(…).light…
```

pytradfri/device.py
```diff
@@ -221,7 +220,7 @@ def set_predefined_color(self, colorname, *, index=0):
             color = COLORS[colorname.lower().replace(" ", "_")]
             return self.set_hex_color(color, index=index)
         except KeyError:
-            _LOGGER.debug('Could not match color name')
+            raise PyTradFriError("Could not match color name in COLOR_NAMES")
```
balloob: Either create a new exception that inherits from this general error (ie `InvalidColor` error) or use the built-in exception for invalid values: `ValueError`.

pytradfri/device.py
```diff
@@ -221,7 +220,8 @@ def set_predefined_color(self, colorname, *, index=0):
             color = COLORS[colorname.lower().replace(" ", "_")]
             return self.set_hex_color(color, index=index)
         except KeyError:
-            _LOGGER.debug('Could not match color name')
+            raise ColorError('Unable to match %s in COLOR_NAMES',
```
balloob: Error message leaks implementation detail. I would go for "Invalid color specified: %s"

ggravlingen:
@balloob Thanks! I'll squash and merge as soon as Travis starts working again and tests pass.

PR #91, opened 2017-10-31
Update LICENSE from opentradfri to pytradfri

schliflo:
As this is now known as pytradfri, the name in the license should be changed accordingly.


ggravlingen:
Yes, must have missed this. Thanks!

PR #98, opened 2017-11-02
Remove references to aiocoap

ggravlingen:


lwis:
We should be referencing tinydtls from https://github.com/flowolf/tinydtls-cython/ or https://pypi.python.org/pypi/DTLSSocket from pypi (they are one and the same).

It's easier if we just say "install the requirements file if you're using asyncio".

ggravlingen:
@lwis I'm reluctant using asyncio, unless within brackets, as it's a quite technical term. Users of pytradfri might not understand the disctinction, hence rewriting it along the lines of "more advanced features such as observe".

lwis:
But observe is available on our sync and async api.

The distinction is sync/async. Users of libraries are typically versed with what they want to do.

ggravlingen:
@lwis gave it another try

lwis:
Looks good to me buddy.

ggravlingen:
@lwis thanks for your feedback!

PR #103, opened 2017-11-05
Simplified color handling (v5.0.0)

ggravlingen:
Nb: breaking changes.

Following the recent firmware update, the gateway now returns more information around colors. This is an attempt to simplify returning what is supported by a bulb. I'm having to go for a breaking change as I need to move methods around in between classes. The color values that are returned seem to be hue, saturation, x and y.

If the user wants to work in the RGB color space, it will have to do the conversion itself. Hence the color example file.

# Change log:

## color.py:
- Added hex to rgb converter
- Added method that returns what features are supported by a bulb (dimmer, hex, xy, mireds)
- Removed abundant methods

## const.py
- Found a set of new constants and added them
- ATTR_LIGHT_COLOR (5706) is hex so I renamed the constant
- Reshuffled constants from other files

## device.py
- Removed inferred methods
- Removed RGB color conversion alltogether
- Only return attributes that are relevant for the bulb (ie a white spectrum bulb does return a color temperature, not a rgb value)
- Reordered setters and getters so that they are in the same order in the different classes

## Tests
- Moved the test bulbs into a separate file and also added motion sensor and dimmer button devices.

## Other:
- Updated tests


pytradfri/device.py
```diff
@@ -257,56 +296,39 @@ def __init__(self, device, index):
         self.index = index
 
     @property
+    def supported_features(self):
+        return supported_color_features(self.raw)
```
ggravlingen: Note to self: rename as light_supported_features

pytradfri/color.py
```diff
@@ -165,3 +179,36 @@ def xy_brightness_to_rgb(vX: float, vY: float, ibrightness: int):
...
     ir, ig, ib = map(lambda x: int(x * 255), [r, g, b])
     return (ir, ig, ib)
+
+
+# https://stackoverflow.com/questions/29643352/converting-hex-to-rgb-value-in-python
+def hex_to_rgb(hex):
+    return tuple(int(hex[i:i+2], 16) for i in (0, 2, 4))
+
+
+def light_supported_features(data):  # what should I name this?
```
lwis: `supported_features`?

pytradfri/device.py
```diff
@@ -223,8 +215,38 @@ def set_predefined_color(self, colorname, *, index=0):
             raise ColorError('Invalid color specified: %s',
                              colorname)
 
-    def set_rgb_color(self, r, g, b, *, index=0):
-        return self.set_values(rgb_to_xyY(r, g, b), index=index)
+    @property
+    def _kelvin_range(self):
```
lwis: Can we not predetermine most of these when we initialise the light?

ggravlingen:
@lwis Thanks for taking the time! I'll go through the comments after work.

Regarding conversion I believe @balloob's view too is to just report back whatever is available in the gateway. I'll fix the method so that it does. Regarding RGB we have to convert but I haven't really figured out what method is used by IKEA yet so those methods will change. Never before realized there were so many ways to represent a color and am happy that I took a class in linear algebra. :smile: 

 I'll update you when I push finalized color commits.

pytradfri/device.py
ggravlingen: Not quite sure I follow your thougts here?

lwis: We could construct this data in an  `__init__`?

ggravlingen: Do you mean to extend __init__ as so?
```
    def __init__(self, device):
        self._device = device

        if not self.can_set_kelvin:
            # White bulb
            self._kelvin_range = (None, None)
        if not self.can_set_color:
            # White spectrum bulb
            self._kelvin_range = (MIN_KELVIN_WS, MAX_KELVIN_WS)
        # Color bulb
        self._kelvin_range = (MIN_KELVIN, MAX_KELVIN)


```

lwis: Something like that, yeah. @property is useful if something is going to change between calls.

ggravlingen: I've made an attempt to move a lot of stuff into __init__ now

pytradfri/color.py
ggravlingen: Changed to supported_features

ggravlingen:
@lwis got a few minutes to check the changes made after your review?

pytradfri/color.py
```diff
@@ -39,32 +45,32 @@
...
 
-def can_kelvin_to_xy(k):
-    return MIN_KELVIN <= k <= MAX_KELVIN
-
-
-# Only used locally to perform normalization of x, y values
-# Scaling to 65535 range and rounding
-def normalize_xy(x, y):
-    return (int(x*65535+0.5), int(y*65535+0.5))
+def kelvin_to_xyY(T, white_spectrum_bulb=False):
```
lwis: Do we still need these conversion methods?

pytradfri/color.py
```diff
@@ -86,82 +92,34 @@ def kelvin_to_xyY(T, white_spectrum_bulb=False):
...
-def rgb2xyzA(r, g, b):
-    # Uses CIE standard illuminant A = 2856K
-    # src: http://www.brucelindbloom.com/index.html?Eqn_RGB_XYZ_Matrix.html
-    # calculation https://gist.github.com/r41d/43e14df2ccaeca56d32796efd6584b48
-    X = 0.76103282*r + 0.29537849*g + 0.04208869*b
-    Y = 0.39240755*r + 0.59075697*g + 0.01683548*b
-    Z = 0.03567341*r + 0.0984595*g + 0.22166709*b
-    return X, Y, Z
+# https://stackoverflow.com/questions/29643352/converting-hex-to-rgb-value-in-python
+def hex_to_rgb(hex):
```
lwis: Do we still need these conversion methods?

pytradfri/device.py
```diff
@@ -202,10 +203,22 @@ def set_dimmer(self, dimmer, *, index=0, transition_time=None):
...
         return self.set_values(values, index=index)
 
+    def set_color_temp(self, color_temp, *, index=0):
+        """Set color temp a light."""
+        if self.can_set_mireds:
+            white_spectrum_bulb = True
+        else:
+            white_spectrum_bulb = False
+
+        values = kelvin_to_xyY(color_temp, white_spectrum_bulb)
```
lwis: Rather than accepting kelvin, can we just accept mireds instead?

ggravlingen: @lwis thanks! Got home a bit to late this evening but will update the code tomorrow.

pytradfri/color.py
ggravlingen: Hmm, not sure why you see this hex_to_rgb method? I don't have it in my local branch, which is in sync with the one on github.

lwis: This isn't in an older commit, are you sure your local copy is in sync with GitHub?

pytradfri/color.py
ggravlingen: No, you are right. I made some tests and it works to just set Mireds. I've removed them.

ggravlingen:
@lwis this should cover your suggestions. Don't know what happened with that commit but now Github should have the most recent version.

pytradfri/device.py
```diff
@@ -136,48 +144,41 @@ class LightControl:
     def __init__(self, device):
         self._device = device
 
-    @property
-    def lights(self):
-        """Return light objects of the light control."""
-        return [Light(self._device, i) for i in range(len(self.raw))]
+        self.can_set_mireds = None
+        self.can_set_color = None
+        self._mireds_range = (None, None)
```
lwis: I'm thinking it may be easier to just set the upper and lower bounds directly, rather than creating an intermediary tuple to store it?

pytradfri/device.py
```diff
@@ -259,57 +268,43 @@ def __init__(self, device, index):
...
+        return supported_features(self.raw)
+
     @property
     def state(self):
         return self.raw.get(ATTR_LIGHT_STATE) == 1
 
     @property
-    def dimmer(self):
-        return self.raw.get(ATTR_LIGHT_DIMMER)
+    def dimmer(self):  # Not convinced this is correct binary calc...
```
lwis: Can you write a test for this if you're unsure? :)

pytradfri/device.py
```diff
@@ -215,8 +219,13 @@ def set_xy_color(self, color_x, color_y, *, index=0):
...
         }, index=index)
 
-    def set_kelvin_color(self, kelvins, *, index=0):
-        return self.set_values(kelvin_to_xyY(kelvins), index=index)
+    def set_hsb(self, hue, saturation, brightness, *, index=0):
+        """Set HSB color settings of the light."""
+        return self.set_values({
+            ATTR_LIGHT_COLOR_SATURATION: hue,
+            ATTR_LIGHT_COLOR_HUE: saturation,
+            ATTR_LIGHT_DIMMER: brightness
```
jurriaan: @ggravlingen  Is it possible to support transition time here as well? I tried setting the transition time, and transitioning between colors works just fine :)

ggravlingen: @jurriaan I was planning on merging this now. Can you raise an issue please and I'll address it separately?

jurriaan: That's okay, will do! :)

Issue #104, opened 2017-11-07
Forgot to change 255 to 254 in color.py?


snillevilla:
### Where are you using pytradfri (eg stand-alone, Home Assistant etc)
Home Assistant, 0.57.2
### Version of pytradfri
The one included in the latest Home Assistant
### Expected behaviour
No errors in log
### Actual behaviour

```
2017-11-07 19:58:53 ERROR (MainThread) [homeassistant.core] Error doing job: Task exception was never retrieved
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/asyncio/tasks.py", line 179, in _step
    result = coro.send(None)
  File "/srv/homeassistant/homeassistant_venv/lib/python3.6/site-packages/pytradfri/api/aiocoap_api.py", line 149, in request
    result = yield from self._execute(api_commands)
  File "/srv/homeassistant/homeassistant_venv/lib/python3.6/site-packages/pytradfri/api/aiocoap_api.py", line 107, in _execute
    yield from self._observe(api_command)
  File "/srv/homeassistant/homeassistant_venv/lib/python3.6/site-packages/pytradfri/api/aiocoap_api.py", line 169, in _observe
    api_command.result = _process_output(r)
  File "/srv/homeassistant/homeassistant_venv/lib/python3.6/site-packages/pytradfri/command.py", line 71, in result
    self._result = self._process_result(value)
  File "/srv/homeassistant/homeassistant_venv/lib/python3.6/site-packages/pytradfri/resource.py", line 46, in observe_callback
    callback(self)
  File "/srv/homeassistant/homeassistant_venv/lib/python3.6/site-packages/homeassistant/components/light/tradfri.py", line 324, in _observe_update
    self._light_data.hex_color_inferred
  File "/srv/homeassistant/homeassistant_venv/lib/python3.6/site-packages/pytradfri/device.py", line 281, in hex_color_inferred
    *xy_brightness_to_rgb(scale(x), scale(y), self.dimmer)
  File "/srv/homeassistant/homeassistant_venv/lib/python3.6/site-packages/pytradfri/color.py", line 142, in xy_brightness_to_rgb
    brightness = ibrightness / 255.
TypeError: unsupported operand type(s) for /: 'NoneType' and 'float'
2017-11-07 19:58:53 WARNING (MainThread) [root] Observation deleted without explicit cancellation
2017-11-07 19:59:00 WARNING (MainThread) [root] Observation deleted without explicit cancellation
```

[...]

PR #105, opened 2017-11-07
Fix brightness after latest update

lwis:
#104


ggravlingen:
This basically makes sure the program doesn't mess up due to user's erroneous input?

lwis:
Yup

ggravlingen:
Thanks! :+1:

PR #107, opened 2017-11-10
Bump to aiocoap alpha

lwis:


ggravlingen:
Release log for reference:
https://github.com/chrysn/aiocoap/blob/master/NEWS

PR #114, opened 2017-11-15
Support and utility for pairing (commissioning)

chrysn:
This adds support for the reverse-engineered (well, just read the network logs) commissioning process as a library call and an example program that spews nice messages to it and crudely displays newly paired devices.

A first commit fixes a race condition that happens when a program (like the example) aggressively starts accessing multiple resources on the server "simultaneously". (Not that the requests would actually overlap because they're CONs and CoAP doesn't by default allow more than one of them at a time, but still, the application can leave that scheduling to the libraries).

### DIFF
--- a/example_pair.py
+++ b/example_pair.py
@@ -1,0 +1,153 @@
+#!/usr/bin/env python3
+"""
+This is an example of how the pytradfri-library can be used to pair new
+devices.
+
+To run the script, do the following:
+$ pip3 install pytradfri
+$ Download this file (example_sync.py)
+$ python3 example_pair.py <IP> <KEY>
+
+Where <IP> is the address to your IKEA gateway and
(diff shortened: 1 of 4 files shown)


ggravlingen:
Thanks for the PR @chrysn! Is this still WIP or ready for review? I'm thinking about the lint errors.

chrysn:
Just went through the lint errors (took me a moment to find which tool the bot uses). If the checks pass this time, it's ready for review.

ggravlingen:
Thanks, the Hound seems pleased! Am a bit curious though about why Travis isn't building this?

ggravlingen:
@chrysn I have released this as 4.1 on pypi. Thank you for your contribution and thank you @lwis for the review! :+1:

PR #134, opened 2018-01-14
Lamp value ranges

### LINKED ISSUE #132: Color ranges
Stand-alone

### Version of pytradfri
5.2.0

I can't find the min and max values of some color properties. I'd like to normalise the values, so I need the absolute minimum and maximum values. Does anyone know them? If they are unkown I'll write a script to find out.

* Hue (best guess: [0, 64800])
* Saturation (best guess: [0, 65279])

Might be a good idea to add things like this to the wiki.

Lakitna:
Closes #132

### DIFF
--- a/pytradfri/const.py
+++ b/pytradfri/const.py
@@ -110,11 +110,8 @@ ATTR_USE_CURRENT_LIGHT_SETTINGS = "9070"
 URL_OTA_FW = "http://fw.ota.homesmart.ikea.net/feed/version_info.json"
 
 
-# Mireds range for which the conversion functions work
-# and that RGB bulbs can show
-RANGE_MIREDS = (40, 600)
-# Mireds range that white-spectrum bulbs can actually show
-RANGE_MIREDS_WS = (250, 454)
+# Mireds range that white-spectrum bulbs can show
+RANGE_MIREDS = (250, 454)
 
 # Hue of a RGB bulb
 RANGE_HUE = (0, 65535)
--- a/pytradfri/const.py
+++ b/pytradfri/const.py
@@ -109,14 +109,25 @@ ATTR_USE_CURRENT_LIGHT_SETTINGS = "9070"
 # URL to json-file containing links to all firmware updates
 URL_OTA_FW = "http://fw.ota.homesmart.ikea.net/feed/version_info.json"
 
+
 # Mireds range for which the conversion functions work
 # and that RGB bulbs can show
-MIN_MIREDS = 40
-MAX_MIREDS = 600
-
+RANGE_MIREDS = (40, 600)
 # Mireds range that white-spectrum bulbs can actually show
-MIN_MIREDS_WS = 250
-MAX_MIREDS_WS = 454
+RANGE_MIREDS_WS = (250, 454)
+
+# Hue of a RGB bulb
+RANGE_HUE = (0, 65535)
+# Effecitive saturation range of a RGB bulb. The bulb will accept
+# slightly higher values, but it won't produce any light.
+RANGE_SATURATION = (0, 65279)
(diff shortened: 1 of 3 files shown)


pytradfri/device.py
```diff
@@ -189,8 +188,9 @@ def set_dimmer(self, dimmer, *, index=0, transition_time=None):
         dimmer: Integer between 0..254
         transition_time: Integer representing tenth of a second (default None)
         """
-        if dimmer < 0 or dimmer > 254:
-            raise ValueError('Dimmer value must be between 0 and 254.')
+        if dimmer < RANGE_BRIGHTNESS[0] or dimmer > RANGE_BRIGHTNESS[1]:
```
Lakitna: Only dimmer values are actively validated. Should we use this opportunity to validate the rest too?

ggravlingen:
@lwis I like these improvements. What are your thoughts?

lwis:
Looks good, and I think doing validation based on the IKEA ranges is a good idea.

Lakitna:
I added value validation to everything with a range. I'll test this in a bit when I have a Gateway present, but I don't expect any problems.

ggravlingen:
Thanks, let me know when to merge!

pytradfri/const.py
```diff
@@ -109,14 +109,25 @@
 # URL to json-file containing links to all firmware updates
 URL_OTA_FW = "http://fw.ota.homesmart.ikea.net/feed/version_info.json"
 
+
 # Mireds range for which the conversion functions work
 # and that RGB bulbs can show
-MIN_MIREDS = 40
-MAX_MIREDS = 600
-
+RANGE_MIREDS = (40, 600)
```
Lakitna: Has anyone successfully set the color temp of a color bulb? The features parameter of color bulbs tells us that you can't and trying it also doesn't do anything.

But at the same time this comment references the ability to set the color temp of a color bulb.

Since I can't set the color temp of a color bulb I'd like to simplify this and only use `RANGE_MIREDS_WS` as `RANGE_MIREDS`. But if someone managed to do this I'd very much like to know how they did it.

ggravlingen: My recollection is that only hex and x/y could be used to change colors. In some cases, the colors were changed on the gateway but nothing happened with the bulbs itself. But it's been a few weeks since I played around with color so I'm 100% sure.

Lakitna: You can also use HSB, but apparently not Mireds. Very annoying, but it is what it is. I'll simplify it like described above.

pytradfri/device.py
```diff
@@ -147,29 +147,6 @@ class LightControl:
...
-        if not self.can_set_color:
-            # White spectrum bulb
-            self._mireds_range = RANGE_MIREDS_WS
-        if self.can_set_color:
-            # Color bulb
-            self._mireds_range = RANGE_MIREDS
-
-        self.min_mireds = self._mireds_range[0]
-        self.max_mireds = self._mireds_range[1]
-
```
Lakitna: I've been quite rigorous here. I don't think it's the job of pytradfri to block sending values to lamps that ignore them.

Lakitna:
If you approve @ggravlingen please feel free to merge :)

ggravlingen:
@Lakitna can you fix the Travis errors before I merge, please?

Lakitna:
Whoops, forgot about them.

Lakitna:
That fixes Travis. Unless you first want to improve Coveralls (and I have no idea what to do there) I suggest merging.

ggravlingen:
To improve coveralls you add unit tests that test the functionality of the classes.

ggravlingen:
Thank you! :taco:

PR #137, opened 2018-01-15
Libcoap observer timeout fix

Lakitna:
I believe it's a known bug (though I can't find the issue) that in sync mode the observer always stops after 90 seconds. It was thought this was a bug in Libcoap itself, but it turns out it's just a default. From `$ coap-client -h`:

```
	-B seconds	break operation after waiting given seconds
			(default is 90)
```

Adding the -B flag with the observation duration to the command fixes this bug. Though make sure `duration > 0` otherwise it goes haywire.

### DIFF
--- a/pytradfri/api/libcoap_api.py
+++ b/pytradfri/api/libcoap_api.py
@@ -104,7 +104,7 @@ class APIFactory:
         path = api_command.path
         duration = api_command.observe_duration
         if duration <= 0:
-            duration = 1
+            raise ValueError("Observation duration has to be greater than 0.")
         url = api_command.url(self._host)
         err_callback = api_command.err_callback
 
--- a/pytradfri/api/libcoap_api.py
+++ b/pytradfri/api/libcoap_api.py
@@ -109,7 +109,7 @@ class APIFactory:
         err_callback = api_command.err_callback
 
         command = (self._base_command('get')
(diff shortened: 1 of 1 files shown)


ggravlingen:
I'd like for @lwis to review any changes to the coap-stuff.

pytradfri/api/libcoap_api.py
```diff
@@ -103,10 +103,13 @@ def _observe(self, api_command):
         """Observe an endpoint."""
         path = api_command.path
         duration = api_command.observe_duration
+        if duration <= 0:
+            duration = 1
```
lwis: Throw a ValueError if it's out of bounds.

Lakitna: I'm hesitant to do that since the default value of duration is 0 in `command.py`. At the same time, I'm hesitant to change that default.

Lakitna: After digging a little deeper I found that default observation duration is 60, not 0. I'll throw the ValueError instead.

Lakitna:
If everybody agrees this can be merged.

ggravlingen:
Thanks for your contributions! :+1:

ggravlingen:
I've pushed 5.2.1 to pypi.

Lakitna:
Awesome!

Issue #138, opened 2018-01-19
Setting color values individually


Lakitna:
### Where are you using pytradfri (eg stand-alone, Home Assistant etc)
Stand-alone

### Version of pytradfri
5.2.1

### Expected behaviour
```
lamp.light_control.set_hsb(1000, 1000, None)
```
Above code should update hue and saturation, but leave brightness unchanged. This is the behaviour of 5.2.0

### Actual behaviour
Program crashes with the following (partial) traceback:

```
  File "/usr/local/lib/python3.6/site-packages/pytradfri/device.py", line 224, in set_hsb
    self._value_validate(brightness, RANGE_BRIGHTNESS, "Brightness")
  File "/usr/local/lib/python3.6/site-packages/pytradfri/device.py", line 266, in _value_validate
    if value < rnge[0] or value > rnge[1]:
TypeError: '<' not supported between instances of 'NoneType' and 'int'
```

In other words. The value validation I wrote blocks this type of use.

### Solution
The easiest solution is changing the value validation to allow `None` as a value. 

See PR #139

ggravlingen:
This can be closed now?

Lakitna:
Yes

PR #139, opened 2018-01-19
Lamp value range validation

Lakitna:
#138

### DIFF
--- a/pytradfri/device.py
+++ b/pytradfri/device.py
@@ -263,7 +263,7 @@ class LightControl:
         """
         Make sure a value is within a given range
         """
-        if value < rnge[0] or value > rnge[1]:
+        if value is not None and (value < rnge[0] or value > rnge[1]):
             raise ValueError('%s value must be between %d and %d.'
                              % (identifier, rnge[0], rnge[1]))
 


pytradfri/device.py
```diff
@@ -263,7 +263,7 @@ def _value_validate(self, value, rnge, identifier="Given"):
         """
         Make sure a value is within a given range
         """
-        if value < rnge[0] or value > rnge[1]:
+        if value is not None and (value < rnge[0] or value > rnge[1]):
```
lwis: `if value and (value < rnge[0] or value > rnge[1]):`

lwis: Could we not have this as;

`if value is None or (value < rnge[0] or value > rnge[1]):`

Lakitna: That would raise the error if `value == None`. Here we never want `None` to raise an error.

lwis:
Is `None` a valid value for all callers of `_value_validate`?

Lakitna:
It should be, but I'll check to be sure.

Lakitna:
Yes, `None` is valid for every instance. It's simply ignored by the lamp without any callback.

ggravlingen:
Nice!

ggravlingen:
@Lakitna do you need this on pypi for your use purposes?

Lakitna:
Scratch that, I made a mistake. Please revert the merge.

This only works for brightness. For every other value it is changed to `0`.

Lakitna:
This makes things a bit more complicated.

I still think it's a good idea to be able to set only a single value, even in compound values (XY, HSB).

I've got an idea, I'll work on it and make a suggestion via a commit in a bit.

PR #140, opened 2018-01-19
Lamp value range validation without mistakes

Lakitna:
I've made a mistake validating what I did in #139.


Lakitna:
There are some lamp values that can be set individually, but others can't.

XY: When setting X you must also set Y, otherwise Y=0
HS: When setting Hue you must also set Saturation, otherwise Saturation=0

When setting either of these 4 values as `None` they will become `0`.

pytradfri/const.py
```diff
@@ -56,8 +56,8 @@
 ATTR_LIGHT_COLOR_HEX = "5706"  # string representing a value in hex
 ATTR_LIGHT_COLOR_X = "5709"
 ATTR_LIGHT_COLOR_Y = "5710"
-ATTR_LIGHT_COLOR_HUE = "5707"
-ATTR_LIGHT_COLOR_SATURATION = "5708"
+ATTR_LIGHT_COLOR_HUE = "5708"
```
ggravlingen: Are you certain about this? Looking at the constants in the Android app, it should be the other way around: https://hastebin.com/zofoqucibe.php

ggravlingen:
@Lakitna can you please also add a few tests for the stuff you've added?

pytradfri/const.py
Lakitna: You are right, I've changed a bit too much here. The other places I flipped hue with sat are correct though.

tests/test_device.py
```diff
@@ -43,3 +44,119 @@ def test_binary_division():
...
+    data = command.data[ATTR_LIGHT_CONTROL][0]
+    assert len(data) is 1
+
+    command = dev.light_control.set_hex_color("4a418a", transition_time=1)
+    data = command.data[ATTR_LIGHT_CONTROL][0]
+    assert len(data) is 2
+
+    command = dev.light_control.set_hex_color("RandomString")
+    data = command.data[ATTR_LIGHT_CONTROL][0]
+    assert len(data) is 1
```
Lakitna: Is it a problem that every string is accepted here? Perhaps we should limit the string length to 6, that could prevent accidental network overload.

ggravlingen: Not needed now, let's do it in another PR late ron.

ggravlingen:
@Lakitna I guess this can be merged now?

Lakitna:
Yes, if you agree it can be merged :)

ggravlingen:
Thank you @Lakitna! Adding those tests is very appreciated. I've pushed this to pypi now.

PR #141, opened 2018-01-19
Adding support for GU10 lights

ggravlingen:
Closes #135

PR #151, opened 2018-01-31
Partial reversal of PR #141

### LINKED ISSUE #150: Remotes recognised as lamps
Stand-alone

### Version of pytradfri
5.3.0

### Expected behaviour
```
light_objects = [dev for dev in devices if dev.has_light_control]
```
Above code returns all lamps, but not other devices

### Actual behaviour
The code returns both lamps and remotes, but not motion sensors.

I this is probably due to the GU10 fix from #135 (PR #141). I think we should reverse this fix.

Lakitna:
PR #141 proved to be based on incorrect assumptions so we're rolling it back.

Closes #150

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ here = path.abspath(path.dirname(__file__))
 with open(path.join(here, 'README.md'), encoding='utf-8') as f:
     long_description = f.read()
 
-VERSION = "5.3.0"
+VERSION = "5.3.1"
 DOWNLOAD_URL = \
     'https://github.com/ggravlingen/pytradfri/archive/{}.zip'.format(VERSION)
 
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -13,7 +13,7 @@
 with open(path.join(here, 'README.md'), encoding='utf-8') as f:
     long_description = f.read()
 
-VERSION = "5.3.0"
+VERSION = "5.3.1"
```
Lakitna: I'm assuming we want to bump the version like this.

ggravlingen: Yes 👍🏻

ggravlingen:
Should be good to merge?

ggravlingen:
Sorry, hit wrong button 🤦‍♂️

Lakitna:
I know that feeling..

I believe it is.

Lakitna:
I know that feeling..

I believe it is.

ggravlingen:
Thank you @Lakitna!

ggravlingen:
Pushed to pypi

PR #155, opened 2018-03-07
Move timeout to I/O API

### LINKED ISSUE #154: How to set timeout for commands?
Currently Command objetcts returned from e.g. `Gateway.get_devices()` have their timeout set to 10 seconds. I would like to set significantly shorter timeouts for certain commands, but this does not seem to be possible with the current API short of modifying `_timeout` of the returned Command directly.

It would be good to have some way to override timeouts. Perhaps on the `request()` method of `APIFactory`?

tazle:
- It turns out that aiocoap doesn't have configurable request timeouts. Need changes there before aiocoap I/O API can have configurable timeouts.

Closes #154

### DIFF
--- a/tests/api/test_libcoap_api.py
+++ b/tests/api/test_libcoap_api.py
@@ -1,8 +1,10 @@
 """Test API utilities."""
+import json
 import pytest
 
 from pytradfri import RequestTimeout
-from pytradfri.api.libcoap_api import retry_timeout
+from pytradfri.api.libcoap_api import APIFactory, retry_timeout
+from pytradfri.gateway import Gateway
 
 
 def test_retry_timeout_passes_args():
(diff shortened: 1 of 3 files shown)


tests/api/test_libcoap_api.py
```diff
@@ -55,3 +57,33 @@ def api(*args, **kwargs):
         retry_api(1, 2, hello='world')
 
     assert len(calls) == 5
+
+
+def test_cosntructor_timeout_passed_to_subprocess(monkeypatch):
```
ggravlingen: Typo?

lwis:
We should use https://github.com/aio-libs/async-timeout for asyncio timeouts.

tests/api/test_libcoap_api.py
tazle: Yes, fixed.

tazle:
I wonder how aiocoap timeouts and async-timeout will interact, and I don't think that can be tested easily. I can try to add support for timeouts using async-timeout, but I don't think the result will be quite optimal.

ggravlingen:
@lwis do you have any input here?

lwis:
Lets leave it as it is for now, if someone wants to add a timeout to a function they can using async-timeout in downstream code.

ggravlingen:
Thank you for your contribution to pytradfri @tazle! 🎉

ggravlingen:
Pushed this to pypi yesterday

PR #161, opened 2018-04-13
Test suite improvement

Lakitna:
I wanted to use newly learned knowledge, so I added and refactored tests.


pytradfri/device.py
```diff
@@ -125,8 +125,6 @@ def power_source(self):
     @property
     def power_source_str(self):
         """String representation of current power source."""
-        if DeviceInfo.ATTR_POWER_SOURCE not in self.raw:
-            return None
```
Lakitna: This part of the code can never be reached, so it should be removed.

ggravlingen:
Great initiative! I'm traveling at the moment but will check asap unless someone beat me to it.

pytradfri/device.py
ggravlingen: This case can never occur?

tests/test_device.py
```diff
@@ -1,26 +1,285 @@
...
+    ATTR_LIGHT_COLOR_X,
+    ATTR_LIGHT_COLOR_Y,
+    ATTR_LIGHT_MIREDS,
+    ATTR_LIGHT_STATE,
+    ATTR_LIGHT_COLOR_HEX
+)
 from pytradfri.device import Device
 from devices import (
-    LIGHT_W, LIGHT_WS, LIGHT_CWS, REMOTE_CONTROL, MOTION_SENSOR)
+    LIGHT_W, LIGHT_WS, LIGHT_CWS, REMOTE_CONTROL)
```
ggravlingen: Did you remove tests for the motion sensor alltogether?

pytradfri/device.py
Lakitna: As far as I was able to determine it always returns `Unkown` from line 130 instead of `None`.

After replacing line 129 with `return 'ERR'` I got the following results:
```python
    dev = Device({})
    print(dev.device_info.power_source_str)  # KeyError: '3' from device.py:140

    dev = Device({'3': {}})
    print(dev.device_info.power_source_str)  # None

    dev = Device({'3': {'6': None}})
    print(dev.device_info.power_source_str)  # 'Unkown'

    dev = Device({'3': {'6': 'abc'}})
    print(dev.device_info.power_source_str)  # 'Unkown'

    dev = Device({'3': {'6': 1}})
    print(dev.device_info.power_source_str)  # 'Internal Battery'

    dev = Device({'3': {'6': True}})
    print(dev.device_info.power_source_str)  # 'Internal Battery'

    dev = Device({'3': {'6': [1, 2]}})
    print(dev.device_info.power_source_str)  # TypeError

    dev = Device({'3': {'6': {1, 2}}})
    print(dev.device_info.power_source_str)  # TypeError
```

`'ERR'` is never a result, thus that spot can't be reached.

tests/test_device.py
Lakitna: I don't think we really need to. There are two device classes (lamps & remotes). These classes are both tested and there are no device-specific cases for the motion sensors like there are for the lamps. The raw JSON outputs for these devices are identically structured.

Raw JSON outputs: https://gist.github.com/Lakitna/b6b041410431bcedff67a947640fb281
Note that the known issue with `3300` and `15009` is still present in this data. I'll fix that later :)

pytradfri/device.py
ggravlingen: This was added by @balloob long backm surely for a reason. It might be that it's a part of the COAP-protocol. I'm not willing to remove the code at this stage.

tests/test_device.py
ggravlingen: But shouldn't we test that the REMOTE_CONTROL and MOTION_SENSOR are both compatible with the code? It won't increase coverage, but ensure that we don't accidentally remove something that creates problem downstream.

pytradfri/device.py
Lakitna: Very well, I'll reverse it. It's not a lot of bloat so it'll be fine.

tests/test_device.py
Lakitna: Hmm you are right. Even though right now they identify identically, that might change in the future. I'll double up here :)

ggravlingen: @Lakitna thanks!

Lakitna:
I might have gone a bit overboard, but pytest still runs in under a second on my pc :)

lwis:
@Lakitna I don't think build times are a concern at the moment :wink:.

Looks like you've got a conflict.

ggravlingen:
Great, thank you for putting them back! Hopefully it pays off as soon as IKEA unlocks data for the other devices. :) I'll merge this as soon as the conflict is resolved.

ggravlingen:
This is great, thank you so much!

PR #164, opened 2018-05-10
Adds function to set multiple properties in the same api call

Hate-Usernames:
Adds a new `set_combined` function that allows both the colour and brightness to be set during the same api call regardless of the colour mode (hs / xy / hex / temp / etc.) used. This was previously only possible with the `set_hsb` function and not any of the other modes.

This allows for smoothly transitioning colour and brightness simultaneously. Example use cases would be sunrise / sunset simulation, switching from bright cool white task lighting to dim warm white relaxation lighting etc.

Important caveat: In my testing this doesn't work with 1st party (IKEA) bulbs, whilst the gateway behaves correctly a physical change to the bulb only occurs for the numerically first parameter, other changed parameters are ignored. It does seem to work for Philips hue bulbs paired with the gateway. For this reason a `can_set_combined` feature flag is included, which is currently only set for Philips bulbs, further testing will be needed to determine exactly which bulbs can and cannot transition different settings simultaneously.

Previous functionality should be unchanged (non-breaking)


lwis:
I'm wondering if we can extend our existing functionality to include
something that allows us to combine calls? So we can essentially have a
little builder for a command.

I'm all for exposing features of the gateway (even if the bulbs don't
support it), but we need to be careful not to create misleading
functionality for unsupported hardware.

Hate-Usernames:
I did consider something like that, but realistically this is only use case I could see (improving transitions when colour and brightness are changed together). A method to join commands seemed less discoverable and we would still need to have a flag which indicates whether the bulb supports this. 

I haven't looked too deeply into the Command class, if the data parameter is always a dict, we could introduce an `__add__` method which merges the data of two commands so that we could then run something like `api(command1 + command2)` But I haven't tested this, so I don't know how it would handle e.g. combining commands sent to different bulbs etc. It's also in my opinion less obvious that this wouldn't be supported by all bulbs.

Other options would be to add brightness to all the colour methods (as in `set_hsb`), but again this is misleading. Or to raise an error if we try to set too many parameters at once on an unsupported bulb, but this isn't currently done anywhere else (e.g. setting the colour on a non-colour bulb).

Happy to take advice as to how to proceed. Admittedly my main goal is to use this in homeassistant, so it doesn't really matter to me how the functionality is exposed :smile:

ggravlingen:
I assume this can be closed now?

Hate-Usernames:
Yes, this use case can now be achieved with #167.

PR #167, opened 2018-05-28
Allow combining commands

Hate-Usernames:
Alternate method to achieve the same result as #164.

For example:
```
command1 = light.light_control.set_dimmer(100, transition_time=20)
command2 = light.light_control.set_color_temp(400)
api(command1 + command2)
```
----

This allows for smoothly transitioning colour and brightness simultaneously. Example use cases would be sunrise / sunset simulation, switching from bright cool white task lighting to dim warm white relaxation lighting etc.

Important caveat: In my testing this doesn't work with 1st party (IKEA) bulbs, whilst the gateway behaves correctly a physical change to the bulb only occurs for the numerically first parameter that is different to the bulb's current state. Other changed parameters are ignored unless the command is repeated. It does seem to work for Philips hue bulbs paired with the gateway. For this reason a `can_combine_commands` feature flag is included, which is currently only set for Philips bulbs, further testing will be needed to determine exactly which bulbs can and cannot transition different settings simultaneously.


setup.py
```diff
@@ -13,7 +13,7 @@
 with open(path.join(here, 'README.md'), encoding='utf-8') as f:
     long_description = f.read()
 
-VERSION = "5.4.2"
+VERSION = "5.5"
```
Hate-Usernames: Not sure how version numbers are being used in this project, assuming we would want a new version if this were to be merged?

lwis:
This code is significantly cleaner, and produces an arguably better API.

ggravlingen:
@Hate-Usernames: thank you for your contribution! Do you need this on pypi instantly?

Hate-Usernames:
At some point I'd like to put these changes into home assistant, which would require a new version pushed to pypi. But I'm unlikely to be able to look at that for at least a week or so - so not instantly but some point soon would be nice! :smile:

PR #169, opened 2018-06-18
Omit main from coverage

Hate-Usernames:
I don't think it makes sense to include the CLI in the coverage stats as it isn't part of the library code, more like an example?


ggravlingen:
Thanks for the PRs! Looks like we need to bump pytest version in the requirements for Travis build to succeed?

Hate-Usernames:
Yes, I think pytest-timeout 1.3.0 requires pytest>=3.6.0 (latest is 3.6.1).
Probably makes sense to bump both versions in requirements_test.txt?

PR #174, opened 2018-06-19
Drop support for Python 3.4

ggravlingen:


ggravlingen:
@lwis probably best if you take a look at this before I merge.

lwis:
If we're going to do this, we may as well clean up the 3.4 code while were there.

ggravlingen:
I haven't yet gotten around to learn the differences of Python 3.4 and its next versions :( Closing this for now.

PR #189, opened 2018-10-02
Catch OSError and socket.gaierror

balloob:
Catch OSError and socket.gaierror when fetching a response with aiocoap.

Can be removed when upstream issue fixed: https://github.com/chrysn/aiocoap/issues/124


ggravlingen:
I'll push to pypi once merged

PR #190, opened 2018-10-06
Add support for using pycoap for communicating with the gateway

moroen:
Added API to use pycoap [](https://github.com/moroen/python-coap-module) as an alternative to coap-client for synchronous communication with the IKEA gateway.

This is intended as a drop-in alternative to coap-client, all existing code should work unchanged except importing the APIFactory:

`from pytradfri.api.pycoap_api import APIFactory`

Admittedly this change is tailored for the use of pytradfri in the IKEA-Tradfri plugin for Domoticz, but hopefully provide some value in other projects as well.

Advantages over coap-client:
- Native python extension module
- Only two dependencies for installation: openssl@1.1 and golang compiler
- Compiles and installs using standard python setuptools

Limitations over coap-client:
- Only works on linux and macOS-versions prior to Mojave (High Sierra and downwards)
- Currently doesn't suppport observations - support for this is planned

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,4 @@
+cython
+aiocoap==0.4a1
+DTLSSocket==0.1.4
+typing>=3,<4
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,0 @@
-aiocoap==0.4a1
-DTLSSocket==0.1.4
-typing>=3,<4
--- a/pytradfri/api/pycoap_api.py
+++ b/pytradfri/api/pycoap_api.py
@@ -16,6 +16,8 @@ _LOGGER = logging.getLogger(__name__)
 CLIENT_ERROR_PREFIX = '4.'
 SERVER_ERROR_PREFIX = '5.'
 
+class ObservationNotImplemented(Exception):
+    pass
 
 class APIFactory:
     def __init__(self, host, psk_id='pytradfri', psk=None, timeout=10):
@@ -50,7 +52,8 @@ class APIFactory:
         """Execute the command."""
 
         if api_command.observe:
-            self._observe(api_command)
+            raise ObservationNotImplemented
+            # self._observe(api_command)
             return
 
         method = api_command.method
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -1,3 +1,4 @@
+cython
```
ggravlingen: Please pin the versoin

ggravlingen:
I don't mind adding support for pycoap, but reckon we should add an extras require to `setup.py`.
```
EXTRAS_REQUIRE = {
    'pycoap': ['cython?', 'golang?', 'pycoap?']
}
```

Also, we should update the readme to indicate support is less developed for pycoap.

requirements.txt
ggravlingen: Oh, and isn't pycoap a requirement?

moroen:
Thank you for the feedback, and thank you for being interested in adding support pycoap. 

I'm working on getting pycoap to work with manylinux in order to publish it on pypi, which  I guess is an requirement for installing it via EXTRAS_REQUIRE in setup.py. As soon as I have this sorted, I'll update the pull request including the required changes in setup.py and readme.

ggravlingen:
Sounds good! Please also add some tests for the api, we want to make sure things are not broken if others contribute to the code.

moroen:
Getting the extension compiled on CentOS 5 for manylinux wheel is turning out to be a total dependency nightmare, I'm ending up with upgrading so many components of centos in the manylinux docker image, that for all practical purposes, it's no longer centos 5...

So I wonder, how much of a problem is it to require pycoap to be installed manually using 
`python3 setup.py install` 

I would argue that it's no more a hassle than compiling coap-client manually, which kind of was the intention...

ggravlingen:
You mean fully leave it to the user to make sure pycoap is working before they install pytradfri?

moroen:
Upgrading to the latest version of openssl has made this extension unworkable in unexpected ways, so at the moment it's "dead in the water". Thank you for your interest in allowing this, I'm closing the pull request, and issue a new one when I have solved the current problem with openssl.

PR #208, opened 2018-11-18
A bit more stable observations

alex3305:
This change fixes a very small bug (I suppose) in the `aiocoap_api` class
where the flag whether to parse the output as json was not there in the
observe method. Hopefully this makes this part of the API a bit more stable.

Also I've added an observation callback as a global variable in the
`ApiResource` class. When tracking this value for an object, it is possible
to invoke the callback manually. This way the observation can also be
invoked when the user makes a manual change. Thus don't having the need to
call out the observation.

Thirdly I've added an explicit error callback to the `ApiResource` observation
which can be used to diagnose issues easily with the use of the print function.

Finally I've also added a check to check whether or not the incoming result has
even changed from the already existing result. Thus eliminating the need to
continiously update the value. This can hopefully make some parts of the
library a little more stable, because it will limit callback calls.

In conclusion, these are very minor changes, but hopefully will make the
library and in turn the applications using this library a little bit more
stable.

### DIFF
--- a/pytradfri/resource.py
+++ b/pytradfri/resource.py
@@ -48,7 +48,6 @@ class ApiResource:
                 callback(self)
 
         def error_callback(value):
-            print(value)
             err_callback(self)
 
         self.observe_callback = observe_callback
--- a/pytradfri/resource.py
+++ b/pytradfri/resource.py
@@ -14,6 +14,7 @@ class ApiResource:
     def __init__(self, raw):
         """Initialize base object."""
         self.raw = raw
+        self.observe_callback = None
 
     @property
     def id(self):
@@ -42,11 +43,17 @@ class ApiResource:
 
             Returns a Command.
             """
-            self.raw = value
-            callback(self)
+            if self.raw != value:
+                self.raw = value
+                callback(self)
 
+        def error_callback(value):
+            print(value)
+            err_callback(self)
+
+        self.observe_callback = observe_callback
         return Command('get', self.path, process_result=observe_callback,
-                       err_callback=err_callback,
+                       err_callback=error_callback,
                        observe=True,
                        observe_duration=duration)
 
@@ -72,4 +79,8 @@ class ApiResource:
         """
         def process_result(result):
             self.raw = result
+
+            if self.observe_callback is not None:
+                self.observe_callback()
+
         return Command('get', self.path, process_result=process_result)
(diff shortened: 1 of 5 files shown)


pytradfri/resource.py
```diff
@@ -42,11 +43,17 @@ def observe_callback(value):
...
             Returns a Command.
             """
-            self.raw = value
-            callback(self)
+            if self.raw != value:
+                self.raw = value
+                callback(self)
 
+        def error_callback(value):
+            print(value)
```
lwis: If the caller is interested in logging, it's their responsibility, not this library.

pytradfri/resource.py
```diff
@@ -14,6 +14,7 @@ class ApiResource:
     def __init__(self, raw):
         """Initialize base object."""
         self.raw = raw
+        self.observe_callback = None
```
lwis: It's not really clear what this is doing, a global observe_callback for all observations?

alex3305: This is only for a specific device/task/group. Not a global for all the objects. At least that's the way I've understand the code so far. Am I wrong about that?

pytradfri/resource.py
alex3305: Amended in cd9f6866456f333a7e3043eb55ef7944d8961292

pytradfri/resource.py
lwis: My bad, misread where this was. Wouldn't it make more sense to call `observe_callback` with the result?

alex3305: Yeah, which is done in `proces_result` function down below. However I guessed it was a good practice to initialise a variable with a default state in the class constructor. That's why it's set to `None` here.

alex3305:
@lwis I've added a cancellation event for observations as per your suggestion. This would make it easier for integration into something like Home Assistant.

Today I've encountered the error I described in that home-assistant/home-assistant#18497 again. The only clue I've got so far is that the API is only returning `decrypt_verify(): found 8 bytes cleartext` and nothing more. I can also still control my lights, but the state is not being reflected in the interface anymore. I am still looking to trace down the root cause of this bug. And I am currently unsure whether it is a Home Assistant or pytradfri issue.

But this change should already (and probably) make somewhat of a difference.

alex3305:
Closing in favor of #214

PR #212, opened 2018-11-27
Resets an observation when a CHANGED is received

alex3305:
It seems that the `pytradfri` lib tends to get stuck on observations for
some people. This change will throw an Error when a CHANGED message is
received. According to some users they only receive these messages when
the observation is in a stuck state.

When implemented correctly - which it is on `example_async.py` - the
observation will restart. I tried this myself by forcing a reset of an
observation after 5 messages and that seems to work beautifully.

Feedback is appreciated!

### DIFF
--- a/pytradfri/api/aiocoap_api.py
+++ b/pytradfri/api/aiocoap_api.py
@@ -145,7 +145,7 @@ class APIFactory:
 
         _, res = await self._get_response(msg)
 
-        api_command.result = _process_output(res, parse_json)
+        api_command.result, _ = _process_output(res, parse_json)
 
         return api_command.result
 
@@ -171,10 +171,12 @@ class APIFactory:
         # Note that this is necessary to start observing
         pr, r = await self._get_response(msg)
 
-        api_command.result = _process_output(r)
+        api_command.result, _ = _process_output(r)
 
         def success_callback(res):
-            api_command.result = _process_output(res)
+            api_command.result, code = _process_output(res)
+            if code == Code.CHANGED:
+                api_command.observe_reset()
 
         def error_callback(ex):
             err_callback(ex)
@@ -223,4 +225,4 @@ def _process_output(res, parse_json=True):
     if not parse_json:
         return output
 
-    return json.loads(output)
+    return json.loads(output), res.code
--- a/pytradfri/command.py
+++ b/pytradfri/command.py
@@ -1,6 +1,7 @@
 """Command implementation."""
 
 from copy import deepcopy
+from .error import ObservationError
 
 
 class Command(object):
@@ -68,6 +69,9 @@ class Command(object):
             self._result = self._process_result(value)
 
         self._raw_result = value
+
+    def observe_reset(self):
+        self.err_callback(ObservationError())
 
     def url(self, host):
         """Generate url for coap client."""
--- a/pytradfri/error.py
+++ b/pytradfri/error.py
@@ -35,3 +35,12 @@ class ServerError(RequestError):
     See section 5.9.3 of draft-ietf-core-coap-04.
     """
     pass
+
+
+class ObservationError():
+    """Error when the observation got stuck.
+
+    This can happen with indefinate observations which can get stuck after
+    some time.
+    """
+    pass


pytradfri/api/aiocoap_api.py
```diff
@@ -145,7 +145,7 @@ def psk(self, value):
 
         _, res = await self._get_response(msg)
 
-        api_command.result = _process_output(res, parse_json)
+        api_command.result, _ = _process_output(res, parse_json)
```
alex3305: Maybe we can check on the `CHANGED` code here too? Although it seems a bit more common than in an observation.

pytradfri/api/aiocoap_api.py
```diff
@@ -171,10 +171,12 @@ def psk(self, value):
         # Note that this is necessary to start observing
         pr, r = await self._get_response(msg)
 
-        api_command.result = _process_output(r)
+        api_command.result, _ = _process_output(r)
```
alex3305: Same here regarding the checking for a `CHANGED` code, but I think it is unecessary here, because we are just starting the observation.

alex3305: After testing for more than a day, adding a check here would be unnecessary.

pytradfri/api/aiocoap_api.py
alex3305: After testing, it seems that we will have to check for the `CHANGED` response here instead of inside an observation. Albeit the `CHANGED` response in an observation callback will occur, it is rather rare, but on this spot it is quite common.

I will ammend my PR with that change in the near future.

alex3305:
I've created a debug version of the above code to test it out on my local (prod) Home Assistant setup. It also includes debug logging which tells me which response was received and in which method/line specifically.

I've come to the following conclusions:
 * When the connection is stuck / hangs / crashed / etc. the response from the gateway or API will always be `CHANGED`. Nothing more, nothing less.
 * The `CHANGED` response will also occur when you try to force a state change upon a light, which already has that state. For instance, turning on a socket switch which already is on.
 * Resetting observations with a simple Error will actually work, but causes quite a bit of logging. It will output like two lines per device, to show that it reset.
 * CPU spikes (which were reported in the Home Assistant issue) are non existent.

I've created a debug version (which I have specified above) which resets all the observation after three successive `CHANGED` responses and that works really well. It kept my connection alive and kicking, and will reset the connection when it has died. \

@lwis @ggravlingen Moving forward I want to ammend this commit with that change, but I am wondering 3 successive `CHANGED` responses is maybe a bit too aggressive? When having quite a lot of devices, this can also occur when starting up the library. It doesn't seem to matter too much though. But maybe 5 would be a better number, what would you think?

lwis:
How often is CHANGED sent in an observation when there's an issue?

alex3305:
@lwis 
> How often is CHANGED sent in an observation when there's an issue?

I haven't noticed one after about 36h. So I'm not sure about that detection. The `CHANGED` is received in the `_request()` method though. With normal use that should only happen once or twice successfully.

lwis:
Ok, I think this is heading in the right direction - I'm just concerned that we may end up resetting observations that are still active.

alex3305:
@lwis I was worried about that too. In my test results I concluded that if one of the observation failed, all of them seem to fail, not only the one. That would also explain the behavior users describe in home-assistant/home-assistant#14386 .

I've found thus far that resetting observations does not have negative side effects at all, except for some additional logging. But I guess I want to up the maximum failed requests to 5 instead of 3, to make it a bit less aggressive for production use.

alex3305:
I'm making very good progress on this. After debugging for a couple of days, I found some very little issues in my code and I'm currently patching those out. I will add another commit when I'm ready with that.

In my logging, it seems that the reset is hit once every six to eight hours or so. It is more prevalent when you operate a light or switch that already has that state. The observation connection doesn't seem to like that very much. Last 24 hours, my observations hit the reset three times. But no CPU spikes or other weird issues and everything keeps working like it should.

I'll keep you posted!

ggravlingen:
Great work so far @alex3305!

PR #214, opened 2018-12-14
Observation reset

alex3305:
This commit resets the observations when they hit a timeout, and also prevents duplicate state changes. Also observations are cancelled whenever possible instead of erroring out.

 1. Whenever the light or socket already contains a specific state (ie. the light is on and is dimmed at 20%), and a change is requested, a filter will be applied and only the changed states will be send to the device.
 2. Observations are reset whenever no state change was received for a limited amount of time. This is dependent on the first change, because Tradfri does not send a successful callback back to us, when no change was applied.
 3. Cancelling observations instead of erroring out will create less error logging. There is still some error logging left, but this is because of the `aiocoap` library.

This was tested for more than a week on my own production installation and the first fix, makes the pytradfri lib about 4 times more stable for me. The second change resets the observations when something bad happens and works wonderfully for me.

There is still something odd happening with observation cancellation which must be errored out instead of cancelled, because otherwise no new observation will be launched by Home Assistant. This is because the `aiocoap` library does not report when a cancellation happened, but it does on errors.

I and some others have tested these changes from: https://github.com/alex3305/hass-components debug and reception has generally been positive.

References:
 * alex3305/hass-components-debug#3
 * home-assistant/home-assistant#9822
 * home-assistant/home-assistant#14386
 * cc. @mitsumaui, @morberg


ggravlingen:
Thank you for your continued effort on this issue!

lwis:
Thanks for the PR, and for continuing to look at stability.

I'm concerned about the second point, and the time you've set - 15 seconds? Do all observations receive a payload within that window? Looking at this PR I fear it may be more appropriate to switch problematic installations to polling rather than aggressively closing sockets.

ggravlingen:
Thank you @alex3305 for addressing the concerns of @lwis. Given the previously closed PRs and your further analysis of the questions that have arisen, it seems there has been comprehensive testing. Albeit for one user only, but that is not anything unique for this PR. Pushing this to Home Assistant's beta branch will quite quickly give us more test user data, while also leaving room to pull the update back.

That said, it's @lwis who's done most of the work on the API and knows what's going on under the hood, so let's hear his views.

Given his approval, please also make sure the tests pass.

alex3305:
@ggravlingen I didn't notice the tests failing. I've fixed that.

I know my previous PR's weren't the best regarding this solution, initially I was looking for a more pragmatic approach. Since than I've moved on and searched for a good and stable solution. Coming up with this one.

I would love to have this integrated into Home Assistant. I had multiple users test this within my own repo and the issue on Home Assistant. See also: 
 * alex3305/hass-components-debug#3
 * home-assistant/home-assistant#14386
 * @morberg, @mitsumaui

ggravlingen:
Thank you! When I talked to Lwis the other day he indicated he's quite busy at the moment but will revert asap. As I mentioned above, for the sake of stability, no changes to the api'a core will be merged without his approval.

alex3305:
Thanks for the feedback. It seems that home-assistant/home-assistant#19222 also had a positive impact on Home Assistant and Tradfri stability. I am currently quite busy myself regarding the holidays. So I understand 😄.

Happy holidays!

pytradfri/api/aiocoap_api.py
```diff
@@ -203,6 +248,14 @@ def error_callback(ex):
 
         return self._psk
 
+    @classmethod
```
lwis: Given that an APIFactory will be created for each gateway, rather than each observation, should this be instanced instead?

pytradfri/api/aiocoap_api.py
```diff
@@ -175,14 +188,46 @@ def psk(self, value):
...
         def error_callback(ex):
             err_callback(ex)
 
         ob = pr.observation
         ob.register_callback(success_callback)
         ob.register_errback(error_callback)
-        self._observations_err_callbacks.append(ob.error)
+        self._observations.append(ob)
+
+    async def _check_observations(self):
```
lwis: Method name needs improving as this isn't just checking for dead observations, please also add a docstring.

pytradfri/api/aiocoap_api.py
```diff
@@ -30,12 +32,16 @@ def _get_psk(self, host, port):
...
 class APIFactory:
+    last_changed = time.time()
+
     def __init__(self, host, psk_id='pytradfri', psk=None, loop=None):
         self._psk = psk
         self._host = host
         self._psk_id = psk_id
         self._loop = loop
-        self._observations_err_callbacks = []
+        self._is_checking = False
```
lwis: Are you expecting a race of some sort? Maybe we should consider using `asyncio.Lock` instead?

pytradfri/const.py
```diff
@@ -129,3 +129,6 @@
 SUPPORT_HEX_COLOR = 4
 SUPPORT_RGB_COLOR = 8
 SUPPORT_XY_COLOR = 16
+
+OBSERVATION_SLEEP_TIME = 5
+OBSERVATION_TIMEOUT = 10
```
lwis: I'm still not sold on these values, if I don't touch any of my lights for a few hours they won't send any events IIRC. Meaning that the `_check_observations` function taking a guess at best, and it'll consider all my observations to be 'dead' when I make a request.

I'm cautious as many users (such as myself) have good stability on observations, I can have connections open for weeks without issue. This change potentially closes observations unnecessarily.

pytradfri/api/aiocoap_api.py
```diff
@@ -175,14 +188,46 @@ def psk(self, value):
...
+                (OBSERVATION_TIMEOUT + OBSERVATION_SLEEP_TIME):
+            _LOGGER.warning('Resetting Tradfri observations...')
+
+            if self._is_resetting:
+                return
+
+            self._is_resetting = True
+
+            while self._observations:
+                ob = self._observations.pop()
```
lwis: Do we need to let aiocoap know that we're ending the observation? If we discard an active observation accidentally does that mean we'll potentially leak a connection?

pytradfri/api/aiocoap_api.py
alex3305: Since this function is called from the callback defined in `_observe()`, when trying to use an instance method, the callback doesn't have any relations to `self` or an instance of `APIFactory` anymore. That's why I put this on a 'global' timer.

If you can point me out on how to use an instance method in the aforementioned callback, I would gladly implement that change.

pytradfri/api/aiocoap_api.py
alex3305: Not specifically a race condition. But because the `asyncio.sleep` function is called within `_remove_timedout_observations`, it can happen that a observation is checked multiple times. This shouldn't cause any problems, but is just unnecessary. Using an `asyncio.Lock` will have the same behaviour, but just seqential instead of parallel.

This check is only done to make sure we check and reset the observations only once at a time, even when controlling multiple lights at once.

pytradfri/api/aiocoap_api.py
alex3305: Ammended in 3fb5c87f231d931ae531b917bc0f1cb132a6f118

pytradfri/const.py
alex3305: That behaviour should never happen though. In my testing it still hasn't. Even after a few hours or close to a day, the observations don't just reset. They just work. What happens is the following flow:

 1. You send a command to the Tradfri gateway
 2. The check method for dead observations is triggered
    1. This method sets checking = True
    2. This method waits for 5 seconds initially
    3. **go to 3 below**
    4. The last updated global timer is checked against the values defined in const
        1. If the time is above the defined times, the observations will be reset
        2. If the time is below the defined times, nothing happens
    5. Set checking = False
 3. The Tradfri hub sends a response in the observation callback
    1. The state of the object is changed from the callback
    2. The last updated global time is updated
    3. **go to 2.4 above**

In my experience most of this happens in a fraction of a second. Almost all of my requests were sub second response times. In some edge cases it took a bit longer, but not that long. We can always up this value though. But I encourage you to test it out for yourself, perhaps with debug logging like:

```python
_LOGGER.debug('Tradfri response_time %f', time.time() - APIFactory.get_last_changed())
```

pytradfri/api/aiocoap_api.py
alex3305: The connections should end as defined in `aiocoap.protocol._BaseClientObservation`. I've reworked this a bit in 17638af0ceb00e40273710bec419086207a3318b so it should also cancel out.

alex3305:
@lwis Any updates?

alex3305:
> @lwis Any updates?

Still working wonderfully for me luckily. Very stable, even with longer transitions and such. I'm hoping that you or @ggravlingen can test this branch for yourself too.

pytradfri/api/aiocoap_api.py
alex3305: `success_callback`  only has it's own scope within the method.

pytradfri/const.py
alex3305: If the device responds, the observation callback will be triggered. It's the same as before.

But if no response from the Tradfri gateway will be received within the timeout window, the observations will be reset. That's the only functional change.

PR #218, opened 2019-01-02
Handle timezone as optional parameter

morberg:
Stealing from https://github.com/home-assistant/home-assistant/blob/dev/script/dev_docker this PR will now make the comments on line 4 and 5 true.

### DIFF
--- a/script/dev_docker
+++ b/script/dev_docker
@@ -19,6 +19,20 @@ case "$MACHINE" in
 esac
 
 docker build -t tradfri-dev -f "$PWD/$DOCKERFILE" .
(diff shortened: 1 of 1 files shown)


ggravlingen:
Thanks for the contribution? How do you execute the script? What params are sent and what do they do?

morberg:
When running `./script/dev_docker` it will behave as before (which does not work on macOS since `/etc/localtime` is not available to docker).

Running `./script/dev_docker Europe/Stockholm` will send the time zone "Europe/Stockholm" to docker instead of linking to `/etc/localtime` which makes it possible to run docker on macOS.

The approach is the same as in Home Assistant's docker script.

ggravlingen:
Thanks!

PR #225, opened 2019-02-16
Switch to instance based health checks, revert dupe checks

lwis:
I've switched the health check to be entirely instance based.

I've removed the dupe checks as I believe this is the responsibility of the caller, if the state in pytradfri is out of date we'll ignore a request that should have gone through, this is also confusing behaviour to the caller.

@alex3305 I would appreciate your feedback on these changes.

Changes are currently untested.


alex3305:
@lwis Thanks for the PR and additional changes. I was very busy lately, but I'm going to review these changes soon.

alex3305:
After some slight modifications I could get this working on my local Home-Assistant install. I will test this the following days with Home-Assistant 0.89.

Thanks again for the input and effort you have put in to this. Great that we can hopefully solve this issue - or make it bearable ;) - together.

alex3305:
Very quick update. On my RasPi 3B+ this change isn't working out at all. The `cancel_failure_timeout()` isn't called, which results in the protocol being reset with every Tradfri request. @lwis @ggravlingen could you both test this and hopefully fix this? I'm not entirely sure what causes this 😞

PR #228, opened 2019-03-14
Fixed _process_output not rasing error when using aiocoap_api

moroen:
Fixes a small error in _process_output.

res.code.is_sucessful is a method, not a property and as such res.code.is_sucessfull will always return true, while res.code.is_successfull() will actually check if the request is successfull or not.


ggravlingen:
Thank you @moroen and @lwis!

PR #232, opened 2019-04-23
Stability improvements

lwis:


balloob:
Going to close this PR as it's old and stale. 

@MartinHjelmare not sure if there is something in this PR that we want to consider picking up.

PR #238, opened 2019-07-16
Add the ability to set color for groups.

### LINKED ISSUE #236: Set color for groups
### Version of  pytradfri 6.0.1

The IKEA app is able to set color for groups, if at least one device in the group supports it.

I suggest to implement the same in  pytradfri. In fact why not just allowing it for all groups?

kjlisby:
I raised issue #236, but was impatient enough to implement the bits I need for myself.

This PR closes #236.


ggravlingen:
Thank you for your contribution @kjlisby! Please check the lint errors and add some tests for the newly added features.

lwis:
If we're going to make the improvement, we should support all that the gateway supports.

kjlisby:
I need help. Am I right that there is no existing tests for groups? It would be easier to extend existing tests. Further, I don’t know how to run the tests locally. Is that described somewhere?

ggravlingen:
One way of running the tests is to run ‘pip install -r requirements_test.txt’ and then run ‘python -m pytest tests’. Another way is to look into tox.

ggravlingen:
You can also run flake locally after installing that requirements file. It’s often quicker than waiting for Travis to complete.

kjlisby:
I tried to do what I could see in the log with apt install and pip install. But I could not make it work. Hence my mis-use of Travis. 

Would you please have a look at the remaining error (something about a coverage tool, I guess). Can that be caused by my doings?

ggravlingen:
> If we're going to make the improvement, we should support all that the gateway supports.

Which other features are you referring to here?

kjlisby:
Color temperature maybe?

ggravlingen:
@kjlisby there was something wrong with the .tox-file, hence the test suite not running correctly. I just fixed this in a separate change.

I haven't looked into groups before so you need to fill me in a bit how they work. If you send a command to a group, are they invoked on all members of the group? What happens if you send a colour command to a bulb that only has a white spectrum?

Also, please bump the version to 6.1.0 here:
https://github.com/ggravlingen/pytradfri/blob/dcd557ea3f06f52ee96c46fb51f73e62750612fa/setup.py#L16

kjlisby:
I don't know any inner workings of the IKEA Trådfri system. 

I do however have a group consisting of two colour bulbs like these: https://www.ikea.com/dk/da/p/tradfri-led-paere-e27-600-lumen-tradlos-kan-daempes-farve-og-hvidt-spektrum-opalhvid-00408612/ and one driver (with a couple of spotlights) as this one https://www.ikea.com/dk/da/p/tradfri-driver-til-tradlos-betjening-gra-50356187/

When I send a set colour command to the group, the driver/spotlight simply ignores the command and the two bulbs are changing colour.

Regarding the change of version number: I will do that tonight. But I actually thought you would prefer to do so yourself.

ggravlingen:
@kjlisby thanks, sounds like it will work as expected then.

I'm ready to merge this when the version has been bumped. Before that can be done, I want to learn what @lwis was referring to above.

lwis:
@ggravlingen @kjlisby this PR only covers adding `set_hex_color` for groups, while I'm suggesting that we support all that [device.py](https://github.com/ggravlingen/pytradfri/blob/master/pytradfri/device.py#L210) supports - this can likely be commonised.

kjlisby:
@lwis and @ggravlingen: I would suggest then to merge the colour setting and then open a new pull request with whatever further extensions you might suggest.

ggravlingen:
@kjlisby @lwis agree with you both., This is a good MVP though.

@kjlisby could I ask you to add the other features in a separate PR? Now that you've got the momentum going. :)

kjlisby:
I have no bulbs supporting color temperature or hue. So I do not fancy much adding those features, since I have no chance of testing them. I think someone owning such bulbs should do that.

I can however add set_xy_color to this PR.

ggravlingen:
OK, that's fair enough. Please add that `set_xy_color` (and tests) and I'll merge the PR and publish on pypi.

ggravlingen:
6,1 is on pypi now. Thank you for your contribution @kjlisby!

PR #240, opened 2019-07-19
Add missing features to group.py

ggravlingen:
This PR adds some missing features that were not added in #238.

As some of the code is now similar in `device.py` and `groups.py` it should be refactored to avoid repetition. This will be done in a separate PR.


pytradfri/group.py
```diff
@@ -46,12 +51,12 @@ def hex_color(self):
     @property
     def member_ids(self):
         """Members of this group."""
-        info = self.raw.get(ATTR_MEMBERS, {})
+        info = self.raw.get(ATTR_GROUP_MEMBERS, {})
 
-        if not info or ROOT_DEVICES2 not in info:
+        if not info or ATTR_HS_LINK not in info:
```
ggravlingen: We already had this attribute stored as `ATTR_HS_LINK` in const.py. Perhaps we should rename it to `GROUP_ROOT_DEVICES`?

PR #260, opened 2019-12-26
Fix mood handling

Rob4t:
Fixes problems with moods as described in the following issues;

https://github.com/ggravlingen/pytradfri/issues/220
https://github.com/ggravlingen/pytradfri/issues/106


pytradfri/__main__.py
```diff
@@ -73,12 +73,6 @@
     else:
         print("No groups found!")
         group = None
-    moods = api(gateway.get_moods())
```
ggravlingen: I'd like to keep this in __main__. Can we add try/catch instead?

Rob4t: Do you expect a flattened list of moods or a dict like:
{
    "Group1": [ <Mood FOCUS> ...]
}
The former would not be very useful without the information for which group the mood is available (since there are no "global" moods)

ggravlingen: I wonder if that's why moods are no longer working, they might have been global when we wrote the code. The purpose of pytradfri is to reflect what data is provided by the gateway without any manipulation. Is the group included in the raw data passed by the gateway?

Rob4t: Unfortunately it is not ( see tests/moods.py ). Also the raw data of a group only has its current active mood_id.
So we have to use group.moods() to fetch the available moods for the given group

ggravlingen: We should only return what the gateway returns, no logic nor manipulation of the data. That has to be done by the consumer in its implementation of pytradfri. That leaves us at option #1 I guess?

Rob4t: Done. Also added the parent_id (group id) to the mood's __repr__

ggravlingen:
Thank you guys! Do you need this on pypi continue working with your implementations?

PR #261, opened 2019-12-26
[./dev_docker] fix TZ issue for docker on mac

xgt001:
When TZ is not explicitly specified, docker on mac throws up this nice error by default:
```
docker: Error response from daemon: Mounts denied:
The path /etc/localtime
is not shared from OS X and is not known to Docker.
You can configure shared paths from Docker -> Preferences... -> File Sharing.
See https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.
```
Refer this upstream docker for-mac upstream issue for details: https://github.com/docker/for-mac/issues/2396
This tiny improvement makes onboarding easier for mac folks who don't want to install libcoap maybe?

### DIFF
--- a/script/dev_docker
+++ b/script/dev_docker
@@ -28,6 +28,15 @@ then
     -v `pwd`:/usr/src/app \
     --rm \
     -t -i tradfri-dev bash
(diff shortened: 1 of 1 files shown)


ggravlingen:
Thank you!

PR #266, opened 2020-02-16
add/remove devices from groups

Rob4t:
This adds functions for adding and removing devices to/from groups. For example it is possible to have one device in multiple groups with those commands. Ikea android/IOS App also seems to not have issues with this.

Caveat: After moving a lightbulb to another group it turns on. (same behaviour with ikea app) Unfortunately this destroys my usecase of being able to dispatch dimmer functions based on the time of the day. (I wanted my bedroom light switch to switch bathroom and bathroom floor lights instead between 0 - 6 o clock).

Anyway, hope someone else finds this usefull...


Rob4t:
Did not see that it was already mentioned here: https://github.com/ggravlingen/pytradfri/issues/211 and here https://github.com/ggravlingen/pytradfri/issues/226

:)

meinarsson:
This would be a create adition to the library. Any chance of it getting merged?

ggravlingen:
I’m truly sorry, this has slipped my mind. Thank you for your contribution @Rob4t!

PR #269, opened 2020-05-19
Update install-coap-client.sh

martijnvwezel:
Fix missing lib issue and asking for the correct permissions.


ggravlingen:
Thank you for you contribution!

PR #271, opened 2020-07-10
don't use fenced code block for inline code

evgeni:
fenced code blocks only really work when they are in an own paragraph, not inline.
if used inline the language selector is rendered as part of the code.


ggravlingen:
Thank you!

PR #274, opened 2020-09-02
Upgrade to aiocoap 0.4b3

### LINKED ISSUE #185: Support for multiple gateways wanted | user in issue willing to donate for a dev to buy a second gateway
#176 discusses that there is no support for multiple gateways in pytradfri. If you're out of things to do, we would appreciate if you could add such support and submit a PR.

balloob:
This upgrades pytradfri to use aiocoap 0.4b3 and migrates credential management to the new approach.

I don't have a Tradfri gateway here so I only got as far as that aiocoap was able to find the credentials. Was not able to verify. 

Fixes #185

Can be tested using:

```
python3 examples/example_async.py <IP> -K <KEY ON BOTTOM OF GATEWAY>
```

This is a breaking change. Async API Factory now needs to be set using `await APIFActory.init(…)`

### DIFF
--- a/pytradfri/api/aiocoap_api.py
+++ b/pytradfri/api/aiocoap_api.py
@@ -7,75 +7,50 @@ import socket
 from aiocoap import Message, Context
 from aiocoap.error import RequestTimedOut, Error, ConstructionRenderableError
 from aiocoap.numbers.codes import Code
-from aiocoap.transports import tinydtls
 
 from ..error import ClientError, ServerError, RequestTimeout
 from ..gateway import Gateway
 
 _LOGGER = logging.getLogger(__name__)
-
-
-class PatchedDTLSSecurityStore:
-    """Patched DTLS store in lieu of a credentials framework.
-       https://github.com/chrysn/aiocoap/issues/97"""
-
-    IDENTITY = None
-    KEY = None
-
-    def _get_psk(self, host, port):
-        return PatchedDTLSSecurityStore.IDENTITY, PatchedDTLSSecurityStore.KEY
-
-
-tinydtls.DTLSSecurityStore = PatchedDTLSSecurityStore
+_SENTINEL = object()
 
 
 class APIFactory:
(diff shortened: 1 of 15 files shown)


balloob:
The other flake8 errors will be squashed once we migrate this repo to black once this is merged.

ludeeus:
Both initial and secondary (without key) seems to work as expected :+1: with my limited test (connect)
```
container# python examples/example_async.py 192.168.2.100 -K Y30VhWBwY5wbuTlR
[]
No lights found!
[<65542 - TRADFRI blind (FYRTUR block-out roller blind)>]
Waiting for observation to end (2 mins)
Try altering any light in the app, and watch the events!
```
```
container# python examples/example_async.py 192.168.2.100
[]
No lights found!
[<65542 - TRADFRI blind (FYRTUR block-out roller blind)>]
Waiting for observation to end (2 mins)
Try altering any light in the app, and watch the events!
```

I don't actually have something paired to it for further testing (the blinds are elsewhere now)

pytradfri/api/aiocoap_api.py
```diff
@@ -7,75 +7,51 @@
...
-    def psk(self, value):
-        self._psk = value
-        PatchedDTLSSecurityStore.KEY = self._psk.encode('utf-8')
-
     async def _get_protocol(self):
         """Get the protocol for the request."""
         if self._protocol is None:
-            self._protocol = asyncio.Task(Context.create_client_context(
-                loop=self._loop))
+            self._protocol = asyncio.Task(Context.create_client_context())
```
MartinHjelmare, line 50: Probably use `asyncio.create_task` instead. We await the task below anyway so it should be the same result.

MartinHjelmare: Or do we need to support older Python versions than Python 3.7?

MartinHjelmare: If so the recommended way is to use `asyncio.ensure_future`.

Issue #237, opened 2019-07-16
Catch aiocoap.error.NetworkError


balloob:
### Version of pytradfri
6.0.1

In #189 we started catching OSError because it could be raised from aiocoap. Aiocoap is starting to catch these errors themselves and wrapping them in `aiocoap.error.NetworkError`. So we need to update our implementation to catch NetworkError besides OSError (not instead off just yet, as not all places have been wrapped yet in aiocoap).

More info https://github.com/chrysn/aiocoap/issues/124

ggravlingen:
To be overly clear, you mean amending this line?
https://github.com/ggravlingen/pytradfri/blob/b63e1d5c79676b7a0b3798bfb4774fb60f51c084/pytradfri/api/aiocoap_api.py#L102

lwis:
Yup, we also need to update our aiocoap (maybe requesting a new PyPi release if there hasn't already been one).

MartinHjelmare:
Closed in #287.

PR #281, opened 2020-09-05
Handle aiocoap shutdown error

MartinHjelmare:
- aiocoap raises a `LibraryShutdown` error when shutting down.
- Handle this error for observations and command calls.
- Make sure we stop what we're doing when getting the shut down error.
- Also handle `CredentialsMissingError` that I've seen raised when loosing connection to the gateway on shutdown.

### DIFF
--- a/pytradfri/api/aiocoap_api.py
+++ b/pytradfri/api/aiocoap_api.py
@@ -5,6 +5,7 @@ import logging
 import socket
 
 from aiocoap import Message, Context
+from aiocoap.credentials import CredentialsMissingError
 from aiocoap.error import (
     RequestTimedOut,
     Error,
@@ -31,6 +32,7 @@ class APIFactory:
         self._observations_err_callbacks = []
         self._protocol = None
         self._reset_lock = asyncio.Lock()
+        self._shutdown = False
 
     @classmethod
     async def init(cls, host, psk_id="pytradfri", psk=None) -> "APIFactory":
@@ -58,6 +60,8 @@ class APIFactory:
         """Reset the protocol if an error occurs."""
         skip = self._reset_lock.locked()
         async with self._reset_lock:
+            if self._shutdown:
+                return
             if skip:
                 # The lock was already acquired, so another task was already
                 # in the process of resetting the protocol, so we don't need
@@ -83,6 +87,7 @@ class APIFactory:
         """Shutdown the API events.
         This should be called before closing the event loop."""
         await self._reset_protocol(exc)
+        self._shutdown = True
 
     async def _get_response(self, msg):
         """Perform the request, get the response."""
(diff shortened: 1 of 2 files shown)


pytradfri/api/aiocoap_api.py
```diff
@@ -26,6 +32,7 @@ def __init__(self, host: str, psk_id="pytradfri", psk=None, internal_create=None
         self._observations_err_callbacks = []
```
MartinHjelmare, line 32: I've tried removing this list, since we don't use it, but then for some reasons the error callbacks are not called by aiocoap. By storing the callbacks aiocoap calls them on shutdown. :man_shrugging:

ggravlingen:
@lwis mind taking a look at this?

ggravlingen:
Could this also cover #237?

MartinHjelmare:
We should replace catching OSError with catching the aiocoap NetworkError. I can make a separate PR with that.

MartinHjelmare:
Although `NetworkError` inherits aiocoap `Error` so we can get away with removing OSError and socket.gaierror.

lwis:
LGTM 👍

I can't remember how self._observations_err_callbacks works without digging, sadly.

ggravlingen:
@MartinHjelmare should we do a release to pypi based on this?

MartinHjelmare:
Yeah, that's a good idea. Thanks!

ggravlingen:
Just tried but something failed. Probably due to missing api token for pypi, which I’m unable to add from my mobile. Will try again tomorrow unless @balloob beats me to it.

On a similar note, should I add more people to pypi admin for this repo?

balloob:
Not necessary. I set it up so if you tag a release it's published

MartinHjelmare:
There was a tag pushed before the release from the release-drafter draft was released. How was the release done?

ggravlingen:
Hmm, I have probably misunderstood something here. I made a mistake and forgot to change the tag before I hit release. Tagged it after the release job and re-ran the action, still with an error.

MartinHjelmare:
Ok. The workflow is normally this:

1. Bump version.
2. Wait until the release-drafter action is finished.
3. Go to releases and edit the draft release, adjust the tag version to be the same as the version bump and release from master branch.
4. The publish action will automatically publish to PyPI.

PyPI doesn't accept two releases with the same version, that's why the second release build failed.

ggravlingen:
Thank you, I haven’t used this before. I’ll make another release.

MartinHjelmare:
There's no need. 7.0.1 is published to PyPI and looks alright.

ggravlingen:
Oh, ok!

MartinHjelmare:
I've removed the tag untagged-5805208d7aabd7ad2870 that was done just before the github release.

PR #289, opened 2020-09-06
Add stale 🤖

ggravlingen:


.github/workflows/stale.yml
```diff
@@ -0,0 +1,19 @@
...
+  schedule:
+  - cron: "30 1 * * *"
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v1
```
MartinHjelmare, line 13: Looks like v3 is the latest major stable release.
https://github.com/actions/stale/releases

MartinHjelmare:
```suggestion
    - uses: actions/stale@v3
```

ggravlingen:
Let’s start with issues only.

MartinHjelmare (approved):

PR #295, opened 2020-11-01
Run unit tests

ggravlingen:
I'm still new to Github action but it doesn't look like we're actually running the test suite?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -31,4 +31,4 @@ jobs:
           black pytradfri examples tests --check
       - name: Run unit tests
         run: |
-          python -m unittest discover tests
+          pytest tests
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,3 +29,6 @@ jobs:
       - name: Check formatting with black
         run: |
           black pytradfri examples tests --check
+      - name: Run unit tests
+        run: |
+          python -m unittest discover tests
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,7 +23,7 @@ jobs:
         run: |
           pip install -r requirements.txt
           pip install -r requirements_test.txt
-          python setup.py install
+          pip install .
       - name: Lint with flake8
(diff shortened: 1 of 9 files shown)


ggravlingen:
Looks like not all tests are run but tests are being returned as passed?

.github/workflows/test.yml
```diff
@@ -29,3 +29,6 @@ jobs:
       - name: Check formatting with black
         run: |
           black pytradfri examples tests --check
+      - name: Run unit tests
+        run: |
+          python -m unittest discover tests
```
MartinHjelmare, line 34: We should use pytest as test runner.

.github/workflows/test.yml
```diff
@@ -29,3 +29,6 @@ jobs:
       - name: Check formatting with black
         run: |
           black pytradfri examples tests --check
+      - name: Run unit tests
+        run: |
+          pytest tests
```
MartinHjelmare, line 34: We need to install the `pytradfri` local package before we run the tests.

ggravlingen: Are you sure about this? Don't think we did this when running unit tests in the previous CI/CD setup.

ggravlingen:
I accidentally commited one of the changes in the test files directly to master rather than to this branch (I'm doing this from the browser). Sorry about that. The same commit has also been done to this PR.

.github/workflows/test.yml
```diff
@@ -23,6 +23,7 @@ jobs:
         run: |
           pip install -r requirements.txt
           pip install -r requirements_test.txt
+          python setup.py install
```
MartinHjelmare, line 26: `pip install .` also works.

.github/workflows/test.yml
MartinHjelmare: The best is to install the package in a virtual environment and run the tests there. `tox` is a tool that does this.

https://docs.pytest.org/en/stable/goodpractices.html

MartinHjelmare (approved):

PR #300, opened 2020-12-15
Reload credentials when the protocol is reset

rubenbe:
Avoid CredentialsMissingError after a coap request fails.

This PR is the result of https://github.com/home-assistant/core/issues/42563
This change has been tested in this context and solves the described issue.

### DIFF
--- a/pytradfri/api/aiocoap_api.py
+++ b/pytradfri/api/aiocoap_api.py
@@ -97,19 +97,23 @@ class APIFactory:
             return pr, r
         except CredentialsMissingError as e:
             await self._reset_protocol(e)
+            await self._update_credentials()
             raise ServerError("There was an error with the request.", e)
         except ConstructionRenderableError as e:
             raise ClientError("There was an error with the request.", e)
         except RequestTimedOut as e:
             await self._reset_protocol(e)
+            await self._update_credentials()
             raise RequestTimeout("Request timed out.", e)
         except LibraryShutdown:
             raise
         except Error as e:
             await self._reset_protocol(e)
+            await self._update_credentials()
             raise ServerError("There was an error with the request.", e)
         except asyncio.CancelledError as e:
             await self._reset_protocol(e)
+            await self._update_credentials()
             raise e
 
     async def _execute(self, api_command):


ggravlingen:
I've followed the discussion for a while and also noted that there have been issues with Tradfri in Home Assistant for quite some time. I believe we should merge it (and then bump in HA core) but would like a second opinion.

@lwis or @MartinHjelmare, what do you think?

ggravlingen (approved):

pytradfri/api/aiocoap_api.py
```diff
@@ -97,19 +97,23 @@ async def _get_response(self, msg):
             return pr, r
         except CredentialsMissingError as e:
             await self._reset_protocol(e)
+            await self._update_credentials()
             raise ServerError("There was an error with the request.", e)
         except ConstructionRenderableError as e:
             raise ClientError("There was an error with the request.", e)
         except RequestTimedOut as e:
             await self._reset_protocol(e)
+            await self._update_credentials()
```
MartinHjelmare, line 106: Why should we update credentials on the other errors besides CredentialsMissingError?

rubenbe: The root cause of the CredentialsMissingError seems to be that after a `_reset_protocol` the credentials class inside libcoap forgets the credentials [0][1]. Users see this problem occur after a network timeout / connection reset.
To resolve this, we pro-actively update the coap credentials.
If we wouldn't, we'd throw the CredentialsMissingError exception after immediately resetting the protocol, reset the protocol again and then only fix the credentials.

[0] https://github.com/chrysn/aiocoap/blob/master/aiocoap/credentials.py#L277
[1] https://github.com/home-assistant/core/issues/42563#issuecomment-725520545
CC @litinoveweedle (original author of the patch)

MartinHjelmare: Unless we're sure that this is required and can reference how this works, I think it's more safe and conservative to only update credentials when we receive that error.

But if we can see good references to the code flow that proofs the logic I can change my mind.

litinoveweedle: @MartinHjelmare hello, I am original author of the patch. I did code analysis of the aiocoap_api.py  and it works like that:

1. in case of any command error error _reset_protocol is called
2. _reset_protocol destroys protocol instance
3. any new following command use _get_protocol to get reference to protocol instance
4. _get_protocol will create new protocol instance if doesn't exists / is empty
5. Unfortunately in this case NO protocol credentials are reinitialized.

Due to _get_protocol being used for many task before and after credentials are initialized (and in fact two different sets of credentials are used) it will require significant code changes to do credentials reinit inside of _get_protocol. But  it is very easy to fix it as proposed:

- when error happens, but protocol is not shutting down (i.e. HA shutting down) simple not only reset BUT also re-init protocol completely, including credentials.
- calling self._update_credentials() after self._reset_protocol(e) does exactly that. So this is not just credentials init, this is in fact whole protocol init after its previous destroy. As we can positively assume, that when we there will be next command following in some time requiring valid credentials. By doing such action, we prevent protocol being reinit only after next command being fired, but in such case without credentials init - and therefore failing to work completely from now on.

I hope that now this PR is more clear. :-)

MartinHjelmare: Thanks for the explanation!

MartinHjelmare (approved):

litinoveweedle:
@rubenbe 
Oops, I've just checked PR diff you prepared (thank you for that), but it seems it misses self._update_credentials() in few other cases. As I explained earlier it seems to be good idea to self._update_credentials() after any _reset_protocol(e), not only after two you selected. Or was there some other reason behind I do not see? Bellow is result of my original patch:

```
    async def _get_response(self, msg):
        """Perform the request, get the response."""
        try:
            protocol = await self._get_protocol()
            pr = protocol.request(msg)
            r = await pr.response
            return pr, r
        except CredentialsMissingError as e:
            await self._reset_protocol(e)
            await self._update_credentials()
            raise ServerError("There was an error with the request.", e)
        except ConstructionRenderableError as e:
            raise ClientError("There was an error with the request.", e)
        except RequestTimedOut as e:
            await self._reset_protocol(e)
            await self._update_credentials()
            raise RequestTimeout("Request timed out.", e)
        except LibraryShutdown:
            raise
        except Error as e:
            await self._reset_protocol(e)
            await self._update_credentials()
            raise ServerError("There was an error with the request.", e)
        except asyncio.CancelledError as e:
            await self._reset_protocol(e)
            await self._update_credentials()
            raise e
```

rubenbe:
@litinoveweedle 
I did take along the entire patch.
Only the github UI makes it confusing because it hides the other two instances below the entire conversation.
See https://github.com/rubenbe/pytradfri/blob/a6297871cd3215d7fbbc664f989a801e69660226/pytradfri/api/aiocoap_api.py#L91

litinoveweedle:
@rubenbe Oh, sorry my mistake, thank you for explanation. That is exactly the reason I am really glad that you prepared the patch and not me. :-) Thank you all for your help.

rubenbe:
Thanks for merging!

@MartinHjelmare @ggravlingen 
Do we tag this fix as a new bugfix release, 7.0.5?
IMHO this change is sufficient to deserve this, but your view on this might be different.

After the release, I can then proceed with a pull request for hass core to have this fix integrated.

ggravlingen:
Yes, please do as 7.0.5!

MartinHjelmare:
Ok, I'll release 7.0.5.

MartinHjelmare:
Released:
https://pypi.org/project/pytradfri/7.0.5/

PR #311, opened 2021-03-24
Cleanup of examples

kiilerix:
Both before and after these changes:

* sync will not receive "observe" events for all devices (something like 5 responses for 10 devices)
* sync seems to receive 3 events on changes: new value, old value, and new value again
* async only receives initial observe events - the observation doesn't really work


MartinHjelmare:
The tests are failing on black code formatter. If we fix that, we should be able to merge here.

PR #312, opened 2021-03-24
Fix observation

### LINKED ISSUE #309: Observe is not reporting

Stand-alone

### Version of pytradfri

7.0.6

### Backend used (aiocoap, libcoap)

aiocoap

### Expected behaviour

State changes of lights, etc. can be observed using example_async.py

### Actual behaviour

Although example_async.py starts and connects to my GW with no problem (no errors or warnings in output and light bulbs are correctly reported), nothing happens when I change the state (e.g. turn on/off) of the light bulbs using either the app or a remote.
When example_async.py starts, it reports the state of the light bulbs correctly and it changes between runs, if I toggle the light in between.

kiilerix:
Fixes #309.

I would suggest considering removing duration completely. It is in general also not sufficient for process handling when using libcoap: users of the api will generally want control when the callback should be cancelled, not specify a timeout upfront.

### DIFF
--- a/pytradfri/api/aiocoap_api.py
+++ b/pytradfri/api/aiocoap_api.py
@@ -175,11 +175,10 @@ class APIFactory:
 
     async def _observe(self, api_command):
         """Observe an endpoint."""
-        duration = api_command.observe_duration
         url = api_command.url(self._host)
         err_callback = api_command.err_callback
 
-        msg = Message(code=Code.GET, uri=url, observe=duration)
+        msg = Message(code=Code.GET, uri=url, observe=0)
 
         # Note that this is necessary to start observing
         pr, r = await self._get_response(msg)
--- a/pytradfri/api/aiocoap_api.py
+++ b/pytradfri/api/aiocoap_api.py
@@ -28,7 +28,7 @@ class APIFactory:
         self._psk = psk
         self._host = host
         self._psk_id = psk_id
-        self._observations_err_callbacks = []
+        self._observations = set()
         self._protocol = None
         self._reset_lock = asyncio.Lock()
         self._shutdown = False
@@ -80,7 +80,7 @@ class APIFactory:
 
             # The error callbacks are called when shutting down the protocol.
             # Clear the saved callbacks
-            self._observations_err_callbacks.clear()
+            self._observations.clear()
 
     async def shutdown(self, exc=None):
         """Shutdown the API events.
@@ -197,7 +197,9 @@ class APIFactory:
         ob = pr.observation
         ob.register_callback(success_callback)
         ob.register_errback(error_callback)
-        self._observations_err_callbacks.append(ob.error)
+        # keep observations alive so they don't get garbage collected
+        self._observations.add(ob)
+        ob.on_cancel(lambda: self._observations.discard(ob))
 
     async def generate_psk(self, security_key):
         """Generate and set a psk from the security key."""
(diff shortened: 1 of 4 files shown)


pytradfri/api/aiocoap_api.py
```diff
@@ -175,11 +175,10 @@ async def request(self, api_commands):
 
     async def _observe(self, api_command):
         """Observe an endpoint."""
-        duration = api_command.observe_duration
         url = api_command.url(self._host)
         err_callback = api_command.err_callback
 
-        msg = Message(code=Code.GET, uri=url, observe=duration)
+        msg = Message(code=Code.GET, uri=url, observe=0)
```
MartinHjelmare, line 181: We're now ignoring the observe duration set on the command. Shouldn't we remove that attribute too? It's already a breaking change by ignoring it.

MartinHjelmare:
It's not clear to me why removing the duration feature is fixing the issue reported.

pytradfri/api/aiocoap_api.py
kiilerix: Yes, I kind of agree that it seems like it would be better to remove duration (observe_duration) completely. But for now, I'm proposing this smaller and safer change.

It seems like this use of duration for aiocoap observe has been wrong since it was introduced in 482dd897. It seems like setting duration to anything but 0 never worked. This change can thus not break anything - only improve things. It doesn't really change the API and I wouldn't call it a breaking change.

The duration parameter is however used for libcoap where it works. But as mentioned in the PR description, I think that could be seen as a misfeature.

MartinHjelmare: What is happening when setting the duration and why doesn't it work as expected?

kiilerix:
Let me try to phrase it differently:

When setting `observe` for a Message, it is used to override the default (which is None) -
https://aiocoap.readthedocs.io/en/latest/module/aiocoap.options.html?highlight=observe#aiocoap.options.Options.observe .
This is passed as a CoAP Header Option where 0 means "add" and 1 means "remove" - https://tools.ietf.org/html/rfc7641#section-2 .

Observe thus only works when `msg.opt.observe` is set to `0`. This field has *nothing* to do with duration. Putting duration in that slot is completely wrong.

The duration feature was actually never implemented for aiocoap, and this change is thus not removing it.

pytradfri/api/aiocoap_api.py
kiilerix: It was putting the duration value in a completely different header field where the only valid values are 0 and 1, and where 0 is the only value that is correct and actually works. When the bogus value is put there, observe doesn't work.

For example, look for `opt.observe == 0` in `aiocoap/protocol.py`.

MartinHjelmare:
Ok. Then I think we should keep duration in the Command api to not break libcoap and just remove it for aiocoap as in your current proposal.

kiilerix:
It is also possible that the observe duration could be implemented in pytradfri for aiocoap. But I think the better API would be to provide a way to cancel a running observe - then it is quite trivial to invoke the cancel after "duration" or other more interactive events. That would/could/should also work for libcoap with the subprocess launched from a thread.

MartinHjelmare:
I think changing the api is a bigger question and should have a separate discussion / PR.

Since this PR is fixing a bug I think it's ok. Let's make sure we document the differences in the command api regarding observation duration in the readme and release notes.

MartinHjelmare:
@kiilerix can you add a paragraph about the observation command api differences in the readme?

PR #314, opened 2021-06-05
Bump aiocoap from 0.4b3 to 0.4.1

rubenbe:
Update aiocoap as there still issues related to tradfri losing its connection occasionally.
I saw the issue re-appear on my own setup, but others have experienced the problem too.
https://github.com/home-assistant/core/issues/42563#issuecomment-849451509

I did a quick sanity check and my lamps still function after upgrading.

Diff: https://github.com/chrysn/aiocoap/compare/0.4b3...0.4.1

PR #332, opened 2021-09-09
Bump version 7.0.7 to use aiocoap version 0.4.1

janiversen:
This seems to solve some of the stability issues, that have been reported.


janiversen:
@MartinHjelmare, my first commit here. Can you please help make this available on pipy.org.

Next step is to add typing to the files, so I can remove cast() in the integration.

MartinHjelmare:
Did you see https://github.com/home-assistant-libs/pytradfri/pull/314?

pytradfri/VERSION
```diff
@@ -1 +1 @@
-7.0.6
+7.0.7
```
MartinHjelmare, line 1: Our standard is to bump the version separately from other changes, just before releasing.

janiversen: ok, I will close this for now and investigate further on my local system. I did now have the problems reported in #314.

PR #333, opened 2021-09-09
Removed false await and return self._protocol directly

janiversen:
We should not await on self._protocol if the task is terminated, that is potentially causing problems.

I will add a couple of test cases in a future PR.

### DIFF
--- a/pytradfri/api/aiocoap_api.py
+++ b/pytradfri/api/aiocoap_api.py
@@ -52,8 +52,9 @@ class APIFactory:
     async def _get_protocol(self):
         """Get the protocol for the request."""
         if self._protocol is None:
-            self._protocol = asyncio.create_task(Context.create_client_context())
-        return await self._protocol
+            task = asyncio.create_task(Context.create_client_context())
+            self._protocol = await task
+        return self._protocol
 
     async def _reset_protocol(self, exc=None):
         """Reset the protocol if an error occurs."""


janiversen:
@MartinHjelmare can we include this in the version bump (release both PRs as one).

MartinHjelmare:
Do you have a issue report or stack trace for this? Would be good to know how the problem manifests so we can recognize what this should fix.

janiversen:
Just look at the code, it is wrong. FIrst time you enter the function async.create_task is called and returns a task object that can be awaited, then it is correctly awaited to get the result.

But second time the function is called, self._protocol already contains the result and therefore should not be awaited. 

In running code the function is called multiple times at least when reconnecting, exactly where we have problems.

MartinHjelmare:
The `self._protocol` attribute has the task, so generally there's no issue with awaiting it multiple times to get the result. There may be issues with this depending on how we handle this attribute in other places, eg when resetting the attribute. That's why it's interesting to see how the issue manifest itself.

janiversen:
Self._protocol only has the task in the first call because that is what asyncio.create_task returns, but that is overwritten in the first await to the result of Context.create_client_context().

So for further calls will no longer be doing an await on the asyncio.create_task, but on the result of Context.create_client_context(), which happens to be awaitable, but clearly that is not the intention of the code.

MartinHjelmare:
There's nothing that overwrites `self._protocol` to the client context after we set the protocol to the task.

PR #334, opened 2021-09-19
Add mypy and fix existing mypy errors

janiversen:
This is the very first step in making the interface typed, allowing a better control in integration tradfri (HA).

It solved this problem:
homeassistant/components/tradfri/link_pytradfri/api/aiocoap_api.py:31: error: Need type annotation for "_observations_err_callbacks" (hint: "_observations_err_callbacks: List[<type>] = ...")  [var-annotated]

This is tested by having a link to the library within the tradfri integration.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@ skip_missing_interpreters = True
 
 [gh-actions]
 python =
-  3.7: py37, lint
+  3.7: py37, lint, typing
   3.8: py38
   3.9: py39
 
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py37, py38, py39, lint
+envlist = py37, py38, py39, lint, typing
 skip_missing_interpreters = True
 
 [gh-actions]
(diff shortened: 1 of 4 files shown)


MartinHjelmare (commented):
We should add a mypy job to our tox file so we check the source when it changes and to know the mypy baseline.

janiversen:
I agree but that is another PR (which I will make later) because that is more than just adding a line to the tox file.

janiversen:
would be nice if you would approve the workflow.

MartinHjelmare (approved):
Thanks!

MartinHjelmare (changes requested):
I think we should actually include the typing env addition to tox in this PR. Otherwise we don't know that mypy runs ok in CI with this fix.

janiversen:
Done, but again I cannot check it (workflow approval).

tox.ini
```diff
@@ -4,7 +4,7 @@ skip_missing_interpreters = True
 
 [gh-actions]
 python =
-  3.7: py37, lint
+  3.7: py37, lint, typing
```
MartinHjelmare, line 7: Please include all the changes from the other PR.

janiversen: ups, I did "git cherry-pick tox" and thought it cherry picked the whole branch, my bad.

Done

MartinHjelmare (approved):
Thanks!

PR #335, opened 2021-09-19
Solve no raw device

janiversen:
Sometimes we receive None in GatewayInfo when running update but only 2-3 times of 10 tests. It seems to be a timing issue either in the gateway or in aiocoap.

This PR does not solve the self.raw problem but secures the library do not crash. There are a number of reported issues in HA which probably will be solved with this PR (and a version bump in HA together with a PR handling the new exception).

### DIFF
--- a/pytradfri/gateway.py
+++ b/pytradfri/gateway.py
@@ -287,7 +287,8 @@ class GatewayInfo:
         def process_result(result):
             self.raw = result
 
-        return Command("get", self.path, process_result=process_result)
+        cmd = Command("get", self.path, process_result=process_result)
+        return cmd if self.raw else None
 
     def __repr__(self):
         return "<GatewayInfo>"
--- a/pytradfri/gateway.py
+++ b/pytradfri/gateway.py
@@ -133,6 +133,8 @@ class Gateway:
         """
 
         def process_result(result):
+            if not result:
+                raise RawError("Raw is None, cannot proceed.")
             return GatewayInfo(result)
 
         return Command(
@@ -228,8 +230,6 @@ class GatewayInfo:
     """This class contains Gateway information."""
 
     def __init__(self, raw):
-        if not raw:
-            raise RawError("Raw is None, cannot proceed.")
         self.raw = raw
 
     @property
--- a/pytradfri/gateway.py
+++ b/pytradfri/gateway.py
@@ -23,6 +23,7 @@ from .const import (
     ATTR_HOMEKIT_ID,
     ATTR_COMMISSIONING_MODE,
 )
+from .error import GatewayInfoError
 from pytradfri.device import Device
 from .group import Group
 from .mood import Mood
@@ -288,7 +289,9 @@ class GatewayInfo:
             self.raw = result
 
         cmd = Command("get", self.path, process_result=process_result)
-        return cmd if self.raw else None
+        if not self.raw:
+            raise GatewayInfoError("Cannot create raw device")
+        return cmd
 
     def __repr__(self):
         return "<GatewayInfo>"
--- a/pytradfri/gateway.py
+++ b/pytradfri/gateway.py
@@ -23,7 +23,7 @@ from .const import (
     ATTR_HOMEKIT_ID,
     ATTR_COMMISSIONING_MODE,
 )
-from .error import GatewayInfoError
+from .error import RawError
 from pytradfri.device import Device
 from .group import Group
 from .mood import Mood
@@ -133,8 +133,6 @@ class Gateway:
         """
 
         def process_result(result):
-            if not result:
-                raise GatewayInfoError("Cannot get gateway info")
             return GatewayInfo(result)
 
         return Command(
@@ -230,6 +228,8 @@ class GatewayInfo:
     """This class contains Gateway information."""
 
     def __init__(self, raw):
+        if not raw:
+            raise RawError("Raw is None, cannot proceed.")
         self.raw = raw
 
     @property
@@ -288,12 +288,11 @@ class GatewayInfo:
         """
 
         def process_result(result):
+            if not result:
+                raise RawError("Raw is None, cannot proceed.")
             self.raw = result
 
-        cmd = Command("get", self.path, process_result=process_result)
-        if not self.raw:
-            raise GatewayInfoError("Cannot create raw device")
-        return cmd
+        return Command("get", self.path, process_result=process_result)
 
     def __repr__(self):
         return "<GatewayInfo>"
--- a/pytradfri/gateway.py
+++ b/pytradfri/gateway.py
@@ -133,6 +133,8 @@ class Gateway:
         """
 
         def process_result(result):
+            if not result:
+                raise GatewayInfoError("Cannot get gateway info")
             return GatewayInfo(result)
 
         return Command(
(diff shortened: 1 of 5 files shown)


pytradfri/group.py
```diff
@@ -40,6 +40,8 @@ def path(self):
     @property
     def state(self):
         """Boolean representing the light state of the group."""
+        if not self.raw:
```
MartinHjelmare, line 43: I don't think this is a correct fix. We rely on the type of `raw` in many places in the code. We need to fix the problem at its source.

janiversen: I agree that it is not solving the problem. But without this patch I need to restart a couple of times each day, and with the patch it has so far been running for around 30hours.

I believe the problem is in aiocoap, but that is not in our scope, and it seems they are very slow in making releases.

janiversen: I think I am able to solve this where self.raw is assigned, which is a far better solution.

```
    def update(self):
        """
        Update the info.

        Returns a Command.
        """

        def process_result(result):
            self.raw = result

        return Command("get", self.path, process_result=process_result)
```

It is clear that Command fails from time to time (especially with many observe calls), and the library do not expect that.

janiversen: I have now implemented the check where self.raw is assigned, I hope you like this patch more.

MartinHjelmare:
Please don't squash or amend commits after review has started to make it easier for readers to track changes. Thanks!

pytradfri/gateway.py
```diff
@@ -287,7 +287,8 @@ def update(self):
         def process_result(result):
             self.raw = result
 
-        return Command("get", self.path, process_result=process_result)
+        cmd = Command("get", self.path, process_result=process_result)
+        return cmd if self.raw else None
```
MartinHjelmare, line 291: The caller of this method expects a Command instance. We can't just change the type. If we don't want to return a Command here it's probably better to raise a custom exception and require the caller to handle that

janiversen: I did change all places inside the library where it is used, but of course there can be / are external usage as well.

I have now added a new custom exception, and changed smart_task (setting of dimmer timer) to handle the exception.

A short life test did not show problems.

janiversen:
> Please don't squash or amend commits after review has started to make it easier for readers to track changes. Thanks!

Sorry my bad, I actually did a completely new commit to remove all the "if not self.raw".

janiversen:
it is btw. interesting that the library partially controls self.raw as in my first version of this PR:
```
    @property
    def has_light_control(self):
        return self.raw is not None and len(self.raw.get(ATTR_LIGHT_CONTROL, "")) > 0
```

janiversen:
Once this is in, I will submit a PR for integration tradfri that catches the exception, and the loop should be closed.

pytradfri/gateway.py
```diff
@@ -287,7 +288,10 @@ def update(self):
         def process_result(result):
             self.raw = result
 
-        return Command("get", self.path, process_result=process_result)
+        cmd = Command("get", self.path, process_result=process_result)
+        if not self.raw:
```
MartinHjelmare, line 292: I still don't think this is the right fix. Why are we checking for `self.raw` here? We don't use `self.raw` in this method. We need to find the source of the problem and address it there.

pytradfri/smart_task.py
```diff
@@ -165,9 +166,14 @@ def set_dimmer_start_time(self, hour, minute):
         """
         #  This is to calculate the difference between local time
         #  and the time in the gateway
-        d1 = self._gateway.get_gateway_info().current_time
-        d2 = dt.utcnow()
-        diff = d1 - d2
+        try:
+            gw = self._gateway.get_gateway_info()
```
MartinHjelmare, line 170: To get the `GatewayInfo` instance we need to use the api and pass the Command to the api and await this. `Gateway.get_gateway_info` returns a `Command` instance, not `GatewayInfo`.

pytradfri/gateway.py
janiversen: There are 2 places in the entire lib where self.raw is assigned, those are now protected. The assignment self.raw = result is the source of the problem. Sometimes a Command fails (that seems to be very common with IKEA Gateway), and that leads to result being None. Now we can either check for self.raw None (like it is done in device) or raise an exception. The library is not prepared for automatic reissue of command, nor do I think it is a good idea, so the best solution is to report the problem to the calling app, and let it decide.

self.raw == None is the problem in a number of reported issues in HA, and I can now produce the problem here relatively easy.

pytradfri/smart_task.py
janiversen: Removed.

pytradfri/gateway.py
MartinHjelmare: But we don't assign `self.raw` here. It's done in the nested function that we define above, which we pass as a callback to be called when processing this command. So checking here doesn't guard anything in this method before we return.

MartinHjelmare: I guess you're talking about the next commit.

MartinHjelmare:
Here's where the `None` command result seems to come from:
https://github.com/home-assistant-libs/pytradfri/blob/d0d19c95e41630346a82d573cef4e79dab6af960/pytradfri/api/aiocoap_api.py#L254-L255

Is there a Command that we expect to return `None` as result, or should we raise already at that point?

There's only four places in the library where we guard `self.raw` for `None` before accessing it.

janiversen:
I believe that  Command can call process_result with None in situations where it is legal. I provoked it by 
1) assign a new device (in my case a bulb), self.raw was set.
2) unscrew the bulb and get color status
3) process_result was called with none, without producing a stack trace.

The 4 places in Device cover (as far as I can see) situations where GatewayInfo.update() have failed to get a result. I do think (but have no proof) that with the RawError they are no longer needed.

janiversen:
Just been thinking this night, the None I saw might be due to a comm. problem I really cannot issue guarantees, but it at least I am not comfortable by moving the raise to such a low level at the moment.

MartinHjelmare:
Yeah. The reported issue stack trace is for a group. So we'd need to guard in more places if we do it like this.

I think we need to dig more and find out what results are expected to be None and when we can raise for missing result and try to do that already in the api module.

The library isn't consistent currently how this is handled. I think that will be apparent when we add more type annotations the code.

janiversen:
I have changed the description (first message), because the PR as is needed for the tradfri integration. Would be nice if we could bump the version or at least have a beta, allowing me to change the integration tradfri.

I am working hard on adding strict typing, and you will see a few PRs soon. I will also continue to research if None is real or an error, in order to do that I intent to extend the tests.

MartinHjelmare:
I'm not convinced that this is the best approach, as I wrote above.

janiversen:
There is a function def retry_timeout(api, retries=3) in libcoapapi.py I wonder why that isn't used.

I cannot see it is used somehow "secretly", but I might have overlooked something.

MartinHjelmare:
I think it's a helper to retry for the applications that use the sync libcoapapi.

janiversen:
Thanks I thought I was overlooking something. I think doing something similar for async is the way to go.

janiversen:
I have decided to write a test that simulates tradfri integration and mock aiocoap. That should allow me test timeout and other “bad weather” conditions.

PR #336, opened 2021-09-19
Add mypy to tox

janiversen:
Once #334 is merged, this PR is ready to be merged.

It adds mypy to tox and adds mypy.ini in the same style as used in HA.

I have tested it locally, but my environment is setup for HA, so I am not sure if it works in a "native" environment. I also cannot test the workflow (as it needs approval).

The idea is to make a number of followup PR with the following priority:
1) type all items used by the tradfri integration.
2) complete typing of all items facing towards the using app
3) have strict typing in the whole library

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@ skip_missing_interpreters = True
 
 [gh-actions]
 python =
-  3.7: py37, lint
+  3.7: py37, lint, typing
   3.8: py38
   3.9: py39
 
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py37, py38, py39, lint
+envlist = py37, py38, py39, lint, typing
 skip_missing_interpreters = True
 
 [gh-actions]
```
MartinHjelmare, line 5: Please add the `typing` env to line 7 below.

janiversen: Done, but please remark I have not been able to test this (workflow is not activated for my account).

janiversen:
Included in #334

PR #338, opened 2021-09-21
Add strict typing part 1

janiversen:
The files are not selected by random, I have made a dependency tree, and the files in this PR depend on nothing or on each other and therefore can be converted without needing to change other files.

The most important file is command.py, with a lot of changes that affect the bulk of the library as well as external use, this file need an extra careful review please.

In order to control that command.py is typed correctly, I have converted around 50% of the library and the findings flowed back into command.py. I have marked a number of code changes for a later PR, since I do not want to mix "themes".

Once this PR is corrected and merged, I will adapt the changes in the next batch and submit it. A batch is defined as:
- files that only depend on each other or already converted files
- the number of changes are below 200
- preferable only one file with complicated conversion

I can easily adapt the batch definition and submit depending on your preference.

### DIFF
--- a/pytradfri/command.py
+++ b/pytradfri/command.py
@@ -7,6 +7,7 @@ from typing import Any, Optional, Callable
 TYPE_PROCESS_RESULT_CB = Optional[Callable[[Any], Optional[str]]]
 TYPE_ERR_CB = Optional[Callable[[str], str]]
 
+
 class Command(object):
     """The object for coap commands."""
 
@@ -42,7 +43,7 @@ class Command(object):
         return self._path
 
     @property
-    def data(self) -> Optional[dict[str,Any]]:
+    def data(self) -> Optional[dict[str, Any]]:
         return self._data
 
     @property
@@ -91,7 +92,9 @@ class Command(object):
         """Generate url for coap client."""
         return "coaps://{}:5684/{}".format(host, self.path_str)
 
-    def _merge(self, a: Optional[dict[str, Any]], b: Optional[dict[str, Any]]) -> Optional[dict[str, Any]]:
+    def _merge(
+        self, a: Optional[dict[str, Any]], b: Optional[dict[str, Any]]
+    ) -> Optional[dict[str, Any]]:
         """Merges a into b."""
         if a is None:
             return b
--- a/pytradfri/command.py
+++ b/pytradfri/command.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 from copy import deepcopy
-from typing import Any, Optional, Callable
+from typing import Any, Callable, Optional
 
 TYPE_PROCESS_RESULT_CB = Optional[Callable[[Any], Optional[str]]]
 TYPE_ERR_CB = Optional[Callable[[str], str]]
@@ -15,7 +15,7 @@ class Command(object):
         self,
         method: str,
         path: list[str],
-        data: Optional[dict[str, Any]] = None,
+        data: dict[str, Any] | None = None,
         *,
         parse_json: bool = True,
         observe: bool = False,
@@ -31,8 +31,8 @@ class Command(object):
         self._err_callback = err_callback
         self._observe = observe
         self._observe_duration = observe_duration
-        self._raw_result: Optional[str] = None
-        self._result: Optional[str] = None
+        self._raw_result: str | None = None
+        self._result: str | None = None
 
     @property
     def method(self) -> str:
@@ -43,7 +43,7 @@ class Command(object):
         return self._path
 
     @property
-    def data(self) -> Optional[dict[str, Any]]:
+    def data(self) -> dict[str, Any] | None:
         return self._data
 
     @property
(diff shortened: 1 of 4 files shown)


janiversen:
@MartinHjelmare all yours :smile:

pytradfri/command.py
```diff
@@ -83,12 +87,16 @@ def path_str(self) -> str:
...
         """Generate url for coap client."""
         return "coaps://{}:5684/{}".format(host, self.path_str)
 
-    def _merge(self, a, b):
+    def _merge(self, a: Optional[dict[str, Any]], b: Optional[dict[str, Any]]) -> Optional[dict[str, Any]]:
         """Merges a into b."""
+        if a is None:
+            return b
+        if b is None:
+            return b
```
KapJI, line 99:
```suggestion
            return a
```

janiversen:
Hmmm seems the githooks needs to be modified, I am so used to having the githook do formatting etc. that I did not think about doing it.

pytradfri/command.py
janiversen: Actually that causes a test to fail (I started by doing that). But anyhow I have changed it to be easier to understand.

janiversen: You are right, but when you look at the callers the parameters are optional and can be None (even though only in error situations), This is one of the cases, where adding typing helped solve a potential bug.

janiversen: I am trying to touch the code as little as possible while adding the typing (and making a note, where to "enhance" the code later). The code is generally written for "good weather", there are a number of places like this (and I have seen enough issues in cases where it was "bad weather").

janiversen:
@KapJI if/when you have time I would really like your input on how solve this:
https://github.com/home-assistant-libs/pytradfri/blob/master/pytradfri/smart_task.py line 37 (definition)
mypy complains:
```
pytradfri/smart_task.py:38: error: Argument 1 to "BitChoices" has incompatible type "Tuple[Tuple[str, str], Tuple[str, str], Tuple[str, str], Tuple[str, str], Tuple[str, str], Tuple[str, str], Tuple[str, str]]"; expected "Tuple[Tuple[str, str]]"  [arg-type]
```
I have defined BitChoise (in util.py) as:
```
class BitChoices(object):
    """Helper class for bitwise dates.

    http://stackoverflow.com/questions/3663898/representing-a-multi-select-field-for-weekdays-in-a-django-model
    """

    def __init__(self, choices: tuple[tuple[str,str]]) -> None:
```
I think my definition is correct, but it WEEKDAY is defined with each entry in the tuple.

Any idea/advice is mostly welcome. Thanks in advance.

KapJI:
Use `...`: `choices: Tuple[Tuple[str,str], ...]`

janiversen:
@KapJI do you have more comments or can you approve the PR ?

pytradfri/command.py
```diff
@@ -1,23 +1,28 @@
...
     """The object for coap commands."""
 
     def __init__(
         self,
-        method,
-        path,
-        data=None,
+        method: str,
+        path: list[str],
+        data: Optional[dict[str, Any]] = None,
```
KapJI, line 18: Don't mix old and new style annotations.
```suggestion
        data: dict[str, Any] | None = None,
```

pytradfri/command.py
```diff
@@ -26,52 +31,52 @@ def __init__(
         self._err_callback = err_callback
         self._observe = observe
         self._observe_duration = observe_duration
-        self._raw_result = None
-        self._result = None
+        self._raw_result: Optional[str] = None
+        self._result: Optional[str] = None
```
KapJI, line 35:
```suggestion
        self._raw_result: str | None = None
        self._result: str | None = None
```

pytradfri/command.py
```diff
@@ -26,52 +31,52 @@ def __init__(
...
         return self._method
 
     @property
-    def path(self):
+    def path(self) -> list[str]:
         return self._path
 
     @property
-    def data(self):
+    def data(self) -> Optional[dict[str, Any]]:
```
KapJI, line 46: You get the idea, there're few more places.
```suggestion
    def data(self) -> dict[str, Any] | None:
```

pytradfri/command.py
```diff
@@ -1,23 +1,28 @@
 """Command implementation."""
+from __future__ import annotations
 
 from copy import deepcopy
+from typing import Any, Optional, Callable
+
+TYPE_PROCESS_RESULT_CB = Optional[Callable[[Any], Optional[str]]]
+TYPE_ERR_CB = Optional[Callable[[str], str]]
```
KapJI, line 8: The old style annotations should be only used here. Until Python 3.10 is a default.

pytradfri/command.py
janiversen: I did not see it as old/new, but you are right they are all changed.

pytradfri/command.py
janiversen: Yup these do not work (yet) with the new style.

KapJI (approved):
LGTM

janiversen:
Thanks.

janiversen:
@MartinHjelmare do you have comments?

MartinHjelmare (approved):
Nice!

PR #339, opened 2021-09-21
Add .pre-commit hooks

janiversen:
This PR adds but do not activate the pre-commit hooks, similar (but reduced) to HA.

It runs black/flake8/isort/mypy

to activate it run:
pre-commit install

However especially black and flake8 made a lot of corrections, so these needs to be merged first, and then githooks will be activated (saving time by not getting surprises from CI).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,14 +14,6 @@ repos:
         name: flake8
         stages: [manual]
 # DO NOT ACTIVATE until all files have been manually updated.
-        additional_dependencies:
-          - pycodestyle==2.7.0
-          - pyflakes==2.3.1
-          - flake8-docstrings==1.6.0
-          - pydocstyle==6.0.0
-          - flake8-comprehensions==3.5.0
-          - flake8-noqa==1.1.0
-          - mccabe==0.6.1
         files: ^(pytradfri|examples|tests)/.+\.py$
   - repo: https://github.com/PyCQA/isort
     rev: 5.9.3
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,66 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 21.7b0
+    hooks:
+      - id: black
+        name: black
+        args:
+          - --safe
+          - --quiet
+        files: ^((pytradfri|examples|tests)/.+)?[^/]+\.py$
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.9.2
+    hooks:
+      - id: flake8
+        name: flake8
+        stages: [manual]
+# DO NOT ACTIVATE until all files have been manually updated.
+        additional_dependencies:
+          - pycodestyle==2.7.0
+          - pyflakes==2.3.1
+          - flake8-docstrings==1.6.0
+          - pydocstyle==6.0.0
+          - flake8-comprehensions==3.5.0
+          - flake8-noqa==1.1.0
+          - mccabe==0.6.1
+        files: ^(pytradfri|examples|tests)/.+\.py$
+  - repo: https://github.com/PyCQA/isort
+    rev: 5.9.3
+    hooks:
+      - id: isort
+        name: isort
+        stages: [manual]
+# DO NOT ACTIVATE until PR 340 have been merged.
+  - repo: local
+    hooks:
+      - id: pylint
+        name: pylint
+        stages: [manual]
+# DO NOT ACTIVATE until all files have been manually updated.
+        entry: pylint
+        language: system
+        types: [python]
+        files: ^(pytradfri|examples|tests)/.+\.py$
+  - repo: local
+    hooks:
+      - id: pytest
+        name: pytest
+        args:
+          - --timeout=30
+          - --cov=pytradfri
+          - --cov-report=html
+        entry: pytest
+        language: system
+        types: [python]
+        files: tests
+  - repo: local
+    hooks:
+      - id: mypy
+        name: mypy
+        entry: ./script/run-pre-commit.sh mypy
+        language: script
+        types: [python]
+        require_serial: true
+#        files: ^(pytradfri|examples|tests)/.+\.py$
+# DO NOT ADD examples/tests until files have been updated.
+        files: ^(pytradfri)/.+\.py$
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,7 +7,6 @@ repos:
         args:
           - --safe
           - --quiet
-        files: ^((pytradfri|examples|tests)/.+)?[^/]+\.py$
   - repo: https://gitlab.com/pycqa/flake8
     rev: 3.9.2
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -49,6 +49,7 @@ repos:
           - --timeout=30
           - --cov=pytradfri
           - --cov-report=html
+        stages: [push,manual]
         entry: pytest
         language: system
         types: [python]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,6 @@ combine_as_imports = true
 [tool.pylint.MASTER]
 py-version = "3.8"
 ignore = [
-    "tests",
 ]
 # Use a conservative default here; 2 should speed up most setups and not hurt
 # any too bad. Override on command line as appropriate.
@@ -30,24 +29,13 @@ load-plugins = [
     "pylint.extensions.code_style",
     "pylint.extensions.typing",
     "pylint_strict_informational",
-    "hass_constructor",
-    "hass_imports",
-    "hass_logger",
 ]
 persistent = false
-extension-pkg-allow-list = [
-    "av.audio.stream",
-    "av.stream",
-    "ciso8601",
-    "cv2",
-]
 
 [tool.pylint.BASIC]
 class-const-naming-style = "any"
 good-names = [
     "_",
-    "ev",
-    "ex",
     "fp",
     "i",
     "id",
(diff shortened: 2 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,41 @@
...
+          - --safe
+          - --quiet
+        files: ^((homeassistant|script|tests)/.+)?[^/]+\.py$
+  - repo: https://gitlab.com/pycqa/flake8
+    rev: 3.9.2
+    hooks:
+      - id: flake8
+        stages: [manual]
+        additional_dependencies:
+          - pycodestyle==2.7.0
```
MartinHjelmare, line 17: Please make sure that the tox run is using the same linting as pre-commit, by running pre-commit in tox.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,41 @@
+repos:
+  - repo: https://github.com/psf/black
+    rev: 21.7b0
+    hooks:
+      - id: black
+        stages: [manual]
+        args:
+          - --safe
+          - --quiet
+        files: ^((homeassistant|script|tests)/.+)?[^/]+\.py$
```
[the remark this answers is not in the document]
janiversen: Yes of course, sorry I forgot to convert this to draft status. I just wanted to signal 2 things:
A) I am working on getting a proper githook, to save work for me and others who commit
B) Testing the code (Black  make changes to a larger number of files).

Before this gets merged, I need to have to code stable in relation to black etc.

.pre-commit-config.yaml
janiversen: That is a very good idea, will do.

.pre-commit-config.yaml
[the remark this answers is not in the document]
janiversen: Done, all are still on manual.

.pre-commit-config.yaml
janiversen: I have to research this a bit more.
- seems flake8 do not run in CI (it comes with a number of “suggestions” when I run it local)
- black runs with —check, where in pre-commit it make changes to the code.

janiversen:
I have now test all entries with:
```
pre-commit run <id> --hook-stage manual --all-files
```
And also
```
pre-commit run --all-files
```
Pytest/black/mypy are active for commit, while flake8/isort/pylint are manual until the sources are updated.

For some reason CI did not like running pre-commit. The different at the moment are:
CI/tox —>  black —check ./
Pre-commit —> black —safe —quiet 
And I would really prefer not to touch tox.ini in this PR.

MartinHjelmare:
It's ok to have black check in CI and write in pre-commit. What's important is that the lint result is exactly the same in both places, ie that we fail linting consistently with exactly the same settings.

Eg, there are plugins used for flake8 in the pre-commit settings. Please check and make sure that they are also applied for tox.

janiversen:
Finally I understood where you were heading, that makes much more sense.

I have add pyproject.toml ("stolen" from HA), and now pylint, black, isort runs identical. flake8 is still missing settings in pyproject.toml (no idea why we do not have them in HA).

I will search for common settings when I start doing the manual changes needed for flake8 (like docstrings).

KapJI:
Flake8 config is in `setup.cfg`: https://github.com/home-assistant/core/blob/dev/setup.cfg
`pyproject.toml` support in flake8 is fairly new but you can start using it there.

janiversen:
thanks a lot, will add it and test it, before somebody approves this PR

janiversen:
Added setup from HA to pyproject.toml, and the number of flagged lines is reduced a lot.

@MartinHjelmare as far as I can see pre-commit and CI find the same problems.

janiversen:
Time to sleep! flake8 was already in setup.cfg (and it does not work yet in pyproject.toml).

KapJI:
I checked again and unfortunately flake8 still doesn't support `pyproject.toml`, it's one of the last tools which doesn't support it yet. So `setup.cfg` is the only option for now.

script/run-pre-commit.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+set -eu
+
+exec "$@"
```
KapJI, line 4: What's the point of this script?

`set -e` just exists on error, but here is one one command anyway
`set -u` errors on unset variables but none of them is used here

janiversen: The point is that without it pre-commit cannot find mypy
```
  entry: mypy
```
fails with "unable to find mypy" (not copied)
whereas 
```
entry: run-pre-commit.sh mypy
```
Works.
I spent a good hour trying to understand why. The other tools works perfect and mypy works perfect on the command line, just not in pre-commit.

janiversen:
Moved black file settings from .pre-commit.yml to pyproject.toml, to have the same setting when committing and running CI.

pyproject.toml
```diff
@@ -0,0 +1,141 @@
...
+    "HomeAssistantError",
+]
+
+[tool.pylint.TYPING]
+runtime-typing = false
+
+[tool.pylint.CODE_STYLE]
+max-line-length-suggestions = 72
+
+[tool.pytest.ini_options]
```
MartinHjelmare, line 134: Is this needed?

pyproject.toml
```diff
@@ -0,0 +1,141 @@
...
+# any too bad. Override on command line as appropriate.
+jobs = 2
+init-hook='from pylint.config.find_default_config_files import find_default_config_files; from pathlib import Path; import sys; sys.path.append(str(Path(Path(list(find_default_config_files())[0]).parent, "pylint/plugins")))'
+load-plugins = [
+    "pylint.extensions.code_style",
+    "pylint.extensions.typing",
+    "pylint_strict_informational",
+    "hass_constructor",
+    "hass_imports",
+    "hass_logger",
```
MartinHjelmare, line 35:
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,141 @@
...
+]
+persistent = false
+extension-pkg-allow-list = [
+    "av.audio.stream",
+    "av.stream",
+    "ciso8601",
+    "cv2",
+]
```
MartinHjelmare, line 43:
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,141 @@
...
+
+[tool.pylint.BASIC]
+class-const-naming-style = "any"
+good-names = [
+    "_",
+    "ev",
```
MartinHjelmare, line 49:
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,141 @@
...
+    "ex",
```
MartinHjelmare, line 50:
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,141 @@
...
+# too-many-ancestors - it's too strict.
+# wrong-import-order - isort guards this
+# consider-using-f-string - str.format sometimes more readable
+# ---
+# Enable once current issues are fixed:
+# consider-using-namedtuple-or-dataclass (Pylint CodeStyle extension)
+# consider-using-assignment-expr (Pylint CodeStyle extension)
+disable = [
+    "format",
+    "abstract-class-little-used",
```
MartinHjelmare, line 81: Please remove the ones we don't need now.

pyproject.toml
```diff
@@ -0,0 +1,141 @@
...
+    "use-symbolic-message-instead",
+]
+
+[tool.pylint.REPORTS]
+score = false
+
+[tool.pylint.TYPECHECK]
+ignored-classes = [
+    "_CountingAttr",  # for attrs
+]
```
MartinHjelmare, line 116:
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,141 @@
...
+
+[tool.pylint.FORMAT]
+expected-line-ending-format = "LF"
+
+[tool.pylint.EXCEPTIONS]
+overgeneral-exceptions = [
+    "BaseException",
+    "Exception",
+    "HomeAssistantError",
```
MartinHjelmare, line 125:
```suggestion
```

MartinHjelmare (commented):
I don't see any changes on what plugins flake8 runs in tox. We need to either remove those plugins from pre-commit or include them in our `requirements_test.txt`.

pyproject.toml
janiversen: Flake8 plugins are moved from .pre-commit-config.yaml to requirements_test.txt

yes, pytest is included .pre-commit-config.yaml and we want it to run identical in tox and pre-commit.

pyproject.toml
janiversen: Done

pyproject.toml
janiversen: Done

pyproject.toml
janiversen: Done but this is a difference to HA.

pyproject.toml
janiversen: Done but this is a difference to HA.

pyproject.toml
janiversen: Done

pyproject.toml
janiversen: done

pyproject.toml
janiversen: I have commented out a few, but that is something I would like to fine-tune when I activate pylint (currently there are pylint errors, which I want to control).

janiversen:
@MartinHjelmare the plugins for flake8 is now loaded in both tox and pre-commit, however that lead to errors in CI, so I had to disable some error codes to have a clean CI.

Changing the code in order to activate the codes again are on the top of my list.

pyproject.toml
MartinHjelmare: We don't really care about what Home Assistant does.

MartinHjelmare:
Please solve the merge conflict.

janiversen:
Done (it was caused because you merged the other PR, thanks for that).

pyproject.toml
```diff
@@ -12,3 +17,107 @@ forced_separate = [
...
+# consider-using-namedtuple-or-dataclass (Pylint CodeStyle extension)
+# consider-using-assignment-expr (Pylint CodeStyle extension)
+disable = [
+    "format",
+#    "abstract-class-little-used",
+    "abstract-method",
+    "cyclic-import",
+    "duplicate-code",
+#    "inconsistent-return-statements",
+#    "locally-disabled",
```
MartinHjelmare, line 74: We'll probably need locally-disabled sooner or later, so that we can disable some warnings locally.

MartinHjelmare (approved):
Looks good!

pyproject.toml
janiversen: OK, I will add that to a future PR.

janiversen: Thanks for merging.

PR #347, opened 2021-09-30
Add strict typing part 2

janiversen:
“Raw” is sadly enough a mixture of different types, which first of all make the typing very large, and secondly not very useable, because different parts of the code depends on a subset of the typing.

Once done with typing, I think it would be good to divide raw in 3-4 different variables, simply to ensure more transparent use.

### DIFF
--- a/pytradfri/device/__init__.py
+++ b/pytradfri/device/__init__.py
@@ -1,11 +1,11 @@
 """Classes to interact with devices."""
 from datetime import datetime
 
-from pytradfri.blind_control import BlindControl
-from pytradfri.light_control import LightControl
+from pytradfri.device.blind_control import BlindControl
+from pytradfri.device.light_control import LightControl
+from pytradfri.device.signal_repeater_control import SignalRepeaterControl
+from pytradfri.device.socket_control import SocketControl
 from pytradfri.resource import ApiResource
-from pytradfri.signal_repeater_control import SignalRepeaterControl
-from pytradfri.socket_control import SocketControl
 
 from ..const import (
     ATTR_APPLICATION_TYPE,
--- a/pytradfri/device/__init__.py
+++ b/pytradfri/device/__init__.py
@@ -1,7 +1,13 @@
 """Classes to interact with devices."""
 from datetime import datetime
 
-from pytradfri.const import (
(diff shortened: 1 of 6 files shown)


pytradfri/resource.py
```diff
@@ -1,49 +1,71 @@
...
+    def __init__(self, raw: TYPE_RAW) -> None:
         """Initialize base object."""
         self.raw = raw
 
     @property
-    def id(self):
+    def id(self) -> str:
         """Id."""
-        return self.raw.get(ATTR_ID)
+        return str(self.raw.get(ATTR_ID))
```
MartinHjelmare, line 31: We don't want to convert `None` to a string if the key is missing in the dict.

janiversen: Corrected. Remark the cast is needed because mypy claims self.raw.get return a RAW_TYPE.

pytradfri/device/__init__.py
```diff
@@ -1,7 +1,13 @@
 """Classes to interact with devices."""
 from datetime import datetime
 
-from pytradfri.const import (
+from pytradfri.device.blind_control import BlindControl
```
MartinHjelmare, line 4: Side note: It would be clearer to have all the names of pytradfri imported relatively to keep them grouped.

pytradfri/resource.py
```diff
@@ -1,49 +1,65 @@
...
+from typing import Any, Dict, List, Union, cast
 
-from .command import Command
+from .command import TYPE_ERR_CB, TYPE_PROCESS_RESULT_CB, Command
 from .const import ATTR_CREATED_AT, ATTR_ID, ATTR_NAME
 
+# type alias
+TYPE_RAW = Dict[str, Union[str, int, List[Dict[str, Union[str, int]]]]]
+TYPE_RAW_SIMPLE = Dict[str, Union[str, int]]
+TYPE_RAW_LIST = Dict[str, List[Dict[str, Union[str, int]]]]
```
MartinHjelmare, line 13: Something to consider for the future is to use a dataclass or TypedDict to represent the RAW data. I'd prefer dataclass without looking into any details yet.

MartinHjelmare (approved):
Looks good!

PR #348, opened 2021-09-30
Use relative import paths for package

janiversen:
Side note from last PR.

VScode complained it did not like the relative imports, but I think I got that solved as well.


MartinHjelmare (approved):
Thanks!

PR #349, opened 2021-10-01
Strictly typing (devices) part 3

janiversen:


pytradfri/device/__init__.py
```diff
@@ -23,76 +26,88 @@ class Device(ApiResource):
     """Base class for devices."""
 
     @property
-    def application_type(self):
+    def application_type(self) -> TYPE_RAW:
         """Return application type."""
-        return self.raw.get(ATTR_APPLICATION_TYPE)
+        return cast(TYPE_RAW, self.raw.get(ATTR_APPLICATION_TYPE))
```
MartinHjelmare, line 31: `None`, which is returned if the key is missing in the dict, is not of `TYPE_RAW`.

pytradfri/device/__init__.py
```diff
@@ -121,48 +136,48 @@ class DeviceInfo:
...
+    def __init__(self, device: Device) -> None:
         """Create object of class."""
         self._device = device
 
     @property
-    def manufacturer(self):
+    def manufacturer(self) -> str:
         """Human readable manufacturer name."""
-        return self.raw.get(DeviceInfo.ATTR_MANUFACTURER)
+        return str(self.raw.get(DeviceInfo.ATTR_MANUFACTURER))
```
MartinHjelmare, line 146: We don't want to convert `None` to a string.

pytradfri/device/light.py
```diff
@@ -24,63 +33,72 @@ class Light:
...
-    def hsb_xy_color(self):
+    def hsb_xy_color(self) -> tuple[int, int, int, int, int]:
         """Return hsb xy color."""
         return (
-            self.raw.get(ATTR_LIGHT_COLOR_HUE),
-            self.raw.get(ATTR_LIGHT_COLOR_SATURATION),
-            self.raw.get(ATTR_LIGHT_DIMMER),
-            self.raw.get(ATTR_LIGHT_COLOR_X),
-            self.raw.get(ATTR_LIGHT_COLOR_Y),
+            cast(int, cast(TYPE_RAW_SIMPLE, self.raw).get(ATTR_LIGHT_COLOR_HUE)),
```
MartinHjelmare, line 86: Why do we change the type of raw here from what's it's typed as below in its definition? That doesn't seem correct.

pytradfri/device/light.py
```diff
@@ -24,63 +33,72 @@ class Light:
...
+            cast(int, cast(TYPE_RAW_SIMPLE, self.raw).get(ATTR_LIGHT_COLOR_SATURATION)),
+            cast(int, cast(TYPE_RAW_SIMPLE, self.raw).get(ATTR_LIGHT_DIMMER)),
```
MartinHjelmare, line 88: Instead of adding all these casts I think we should try to type raw better from the beginning. I think we should start in that end.

janiversen: I would love to do that, the definition (TYPE_RAW) is in resource.py

I have tried many times to limit the definition, but raw contains nearly anything.

I created TYPE_RAW_SIMPLE and TYPE_RAW__LIST as subtypes of TYPE_RAW, this needed to be able type a lot of usages e.g var = self.raw[ATTR_VAR][1], which is totally legal.

I think the right way to go is to fight through all the casts, then we have the fundament to convert raw into a proper dataclass.

MartinHjelmare: Why is that better? I don't think we gain more knowledge about raw by doing this. We should start typing at the center and move outwards.

janiversen: ?? the center is typed (resource.py), I do work my way from the inside and out.

Again the problem is the all the different types of data all stored in raw. But if you have a better suggestion on to type raw I am all ears.

It is better to finalize strict typing, because then it is a lot clearer how exactly raw is used. Remember the usage is often indirected through returns in 2-3 functions.

MartinHjelmare: Raw is the response data from the gateway so that's even more central. We should type it according to what that response is not by how we use it.

janiversen: And that is what I have done in an earlier PR, the definition is in resource.py (TYPE_RAW)

pytradfri/device/__init__.py
janiversen: Corrected.

pytradfri/device/__init__.py
janiversen: Corrected.

pytradfri/device/light.py
janiversen: This is needed, TYPE_RAW consist of 2 subtypes
- TYPE_RAW_SIMPLE, which means the dict contains simple type str, int, None
- TYPE_RAW_LIST, which means the dict contains a list.
I decided to make these 2 subtypes as their own types, in order to quite a number of casts.

But it is actually no longer needed here.

janiversen:
Now I just need to get rid of “type: ignore” which is not permitted (in my mind).

janiversen:
FINALLY, raw definitively needs to be a dataclass, but the last commits are a good example, why I need to make it strictly typed first, so I can see how it is used (DO NOT forget this is reverse engineering, there are no documentation on what the device returns).

In parallel to the tests in the code, I test with a live gateway.

MartinHjelmare:
I think we should start by collecting examples of all command responses and try to use pydantic to define the response types. Then the rest of the library will be much easier to type after that.

https://pydantic-docs.helpmanual.io/

We already have some examples here:
https://github.com/home-assistant-libs/pytradfri/blob/master/tests/devices.py

janiversen:
Thanks for the link, I had a look at the documentation, seems pydantic in its base form is something like an intelligent dataclass, but used correctly it can save a lot of typing.

You say it will be easier to type, but in reality we change raw to be a data class, so it is far more than just typing and it will be a very big change in the library, since it will affect 2/3 of the code, and to a large extent it is something that needs to be done in 1 PR.

Before doing that, I think it is worthwhile to separate the layers (api should only depend on itself, and the same for device). Doing that is correct but will also help with the implementation of a new raw.

I will park this PR for a while.

PR #357, opened 2021-10-11
Remove flake8 plugins

janiversen:
Replaces the automatic ones.

Needs to be done together due to inter-dependencies.


janiversen:
It turns out not to be possible to bump flake8 to 4.0.0 due to especially flake8-comprehensions and flake8-noqa which both require flake8 < 4.

PR #358, opened 2021-10-12
Bumb version to 7.0.7

janiversen:
Can we please release version 7.0.7

Line 241-243 in pytradfri/api/aiocoap_api.py solved a stack trace, when the device is already paired.

please see
https://github.com/home-assistant/core/pull/57524

@MartinHjelmare you offered earlier to help make a new release and upload to pypi.org, I would like to take you up on that offer.

Please merge #359, before making a release, so we have a clean slade.


janiversen:
Thanks a lot.

PR #378, opened 2021-11-07
Add timeout to async version

janiversen:
Both pytradfri as well as the tradfri integration in HA have for a longer time had a problem that requests stacked whenever there was a network problem.

The problem is that the async lib aiocoap do NOT raise timeout, and thus the requests hangs forever.

This PR adds a “wait_for” at the call to aiocoap, and this (re)introduces the timeout. 

The timeout is made available at async APIFactory.request() as an optional parameter “timeout=“.

Once this PR is approved and merged, a release is needed so it can be made generally avialable, a separate PR is made for the version bump.

### DIFF
--- a/pytradfri/api/aiocoap_api.py
+++ b/pytradfri/api/aiocoap_api.py
@@ -98,12 +98,12 @@ class APIFactory:
         await self._reset_protocol(exc)
         self._shutdown = True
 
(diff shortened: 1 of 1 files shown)


MartinHjelmare (approved):
Looks good!

PR #379, opened 2021-11-07
Bump version to 7.2.0

janiversen:
This PR follow #378.

The version is changed to 7.2.0 because the API is change in a NON-breaking way.

async APIFactory.request() has an optional parameter “timeout=“.
Default is None, meaning wait forever (like in the current version)
timeout=<seconds> to give the max. wait time.


MartinHjelmare (approved):
Thanks!

MartinHjelmare:
https://pypi.org/project/pytradfri/7.2.0/

PR #380, opened 2021-11-09
Update author

janiversen:
Not a big deal,

but I think that being the third overall most active contributor (https://github.com/home-assistant-libs/pytradfri/graphs/contributors ) makes this PR fair.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ setup(
     description="IKEA Trådfri/Tradfri API. Control and observe your "
     "lights from Python.",
     long_description=LONG_DESCRIPTION,
-    author="balloob, lwis, ggravlingen",
+    author="balloob, lwis, ggravlingen, janiversen",
     author_email="no@email.com",
     long_description_content_type="text/markdown",
     url=GITHUB_URL,


ggravlingen (approved):

ggravlingen:
Definitely agree, thanks for all your improvements to the library!

janiversen:
Thanks, more to come, (like converting raw to a dataclass), but I miss a place to get feedback on my ideas. Do you think making such “discussions” could be done in an issue ?

ggravlingen:
I took the liberty of adding you as a maintainer to Pypi as well.

ggravlingen:
I've seen there's a discussions feature here in Github. Maybe we should introduce that?

janiversen:
To be honest I have not tried that, but I am totally open….important for me is to get a second opinion on basic changes, before I implement them.

ggravlingen:
This library is part of the HA echosystem so maybe @MartinHjelmare can provide some insight on how this is normally done in the HA organisation?

janiversen:
Yes it would be nice to hear the opinion of @MartinHjelmare 

I just looked at the documentation:
https://docs.github.com/en/discussions/quickstart#enabling-github-discussions-on-your-repository
and (again( it demands superpowers, which I assume you have.

MartinHjelmare:
We think it's ok to just keep using issues for discussion in this repo. In lower traffic repos activating discussions may cause more confusion than help.

ggravlingen:
Thanks for the clarification!

PR #387, opened 2021-11-21
Clean and fix tests

MartinHjelmare:
- Fix incorrect tests that didn't run correctly due to incorrect logic in the tests.
- Make sure to copy constant fixture data before modifying it in tests and do it before instantiating test targets.
- Add missing test package module.
- Ignore mypy errors for tests.


ggravlingen (approved):
Looks good

PR #388, opened 2021-11-21
Type aiocoap api

MartinHjelmare:


pytradfri/command.py
```diff
@@ -32,8 +32,8 @@ def __init__(
         self._err_callback = err_callback
         self._observe = observe
         self._observe_duration = observe_duration
-        self._raw_result: str | None = None
-        self._result: str | None = None
+        self._raw_result: dict | str | None = None
```
ggravlingen, line 35: When is it a `dict` and when is it a `str`? Should we perhaps add a line or two to document this?

pytradfri/api/aiocoap_api.py
```diff
@@ -218,21 +232,26 @@ async def _observe(self, api_command, timeout):
...
                 return
-            err_callback(exc)
+
+            if err_callback:
+                err_callback(exc)
 
         observation = pr_req.observation
+        # The observation is set on the request
+        # since we pass a Message with observe set above.
+        assert observation is not None
```
ggravlingen, line 249: I'm using Bandit for another project (and I've seen core uses this too). It throws a warning when using `assert` so should we consider raising instead?
https://bandit.readthedocs.io/en/latest/plugins/b101_assert_used.html

pytradfri/api/aiocoap_api.py
```diff
@@ -15,55 +17,74 @@
     RequestTimedOut,
 )
 from aiocoap.numbers.codes import Code
+from aiocoap.protocol import BlockwiseRequest
+
+from pytradfri.command import Command
```
ggravlingen, line 22: Should we move this to a relative import?

ggravlingen (commented):
Looks good but I have a few questions.

pytradfri/command.py
MartinHjelmare: Yeah. It's a string when generating the psk and when getting the endpoints since those need custom parsing. I can add a comment.

I have follow up work planned that will narrow down the result types.

pytradfri/api/aiocoap_api.py
MartinHjelmare: We use assert here to tell mypy that we know that the observation is not None. It's not used during runtime.

pytradfri/api/aiocoap_api.py
MartinHjelmare: Yes, that should work and is more consistent.

pytradfri/command.py
janiversen: It can also be a list[dict].

pytradfri/api/aiocoap_api.py
janiversen: Having “assert” in production code in not a good idea…it is not used for the runtime logic, but the assert is executed.

pytradfri/command.py
MartinHjelmare: Are you referring to when multiple commands are sent? Each command result is processed individually.

janiversen: I am referring to what I have seen (in debugger) being returned from aiocoap and assigned.

janiversen: I believe the best way to reproduce it, is to make several events in the gateway at the same time using e.g. 2 ikea buttons and 2 lights. I do not believe you can produce it by sending commands.

MartinHjelmare: Yes, you're right, the command to get devices, groups etc, returns a list. I'll fix this.

pytradfri/api/aiocoap_api.py
MartinHjelmare: It's executed if it's not turned off, yes. We can check and raise some error instead. I just want to make clear that we don't do this for any runtime purposes. It's just to satisfy the static check of mypy.

If we would be wrong about our assumptions it's a bug and we would raise anyway below.

pytradfri/command.py
janiversen: when correcting please remember it can also be an empty list.

PR #398, opened 2021-12-22
fix if-expression for user-id check

4ch1m:
Hi there,

If run via "/bin/sh" (instead of "/bin/bash") the first if-expression of that script will always fail with...

```
./install-coap-client.sh: 3: [: Illegal number: 
```

... because you're comparing a string with an integer.

`$(id -u)` should be used instead.

Thanks.


ggravlingen:
Thanks!

PR #400, opened 2021-12-28
Refactor air purifier

ggravlingen:
I played around a bit with my air purifier and noted I was able to set the fan speed in increments of 1, starting from the value 2 and up to and including the value 50. This PR adds support for setting a fan speed in that range.

I also propose some breaking name changes of the API for the air purifier.


pytradfri/device/air_purifier.py
```diff
@@ -35,25 +36,25 @@ def raw(self) -> TypeRaw:
...
+        Return auto mode on or off.
+
+        Auto mode sets the fan speed automatically based on the air quality.
+        """
+        return self.raw[ATTR_AIR_PURIFIER_MODE] == ATTR_AIR_PURIFIER_MODE_AUTO
+
+    @property
+    def state(self) -> bool:
+        """Return device state, ie on or off."""
+        return cast(int, self.raw[ATTR_AIR_PURIFIER_MODE]) > 0
```
MartinHjelmare, line 50: Is it never interesting to know the mode above 1?

ggravlingen:
AE6B6F7A-EF1B-44FC-85D8-FF1E85641311

pytradfri/device/air_purifier.py
ggravlingen: Please see the screen shot from the official app below. The way I see it, there is one mode (auto). Alternatively, the user sets the fan speed using the slider.

MartinHjelmare: Ok, but we're now hiding this api info, making it possible to access it, right?

ggravlingen: We expose auto mode, we expose on/off and we expose the fan speed. Shouldn’t that cover all use cases? Using the previous “mode” only set the fan speed to some fixed values.

Mariusthvdb: Thanks for developing this. Talking about exposing, would exposing air quality be possible? If an Fr for that is needed I d be glad to write that.

ggravlingen:
@Mariusthvdb it's already exposed here: https://github.com/home-assistant-libs/pytradfri/blob/99c207dd92a1affc6f49307e04056c25b73f9cb8/pytradfri/device/air_purifier.py#L71

You probably want it in Home Assistant and for that, the Tradfri-integration would have to be completely rewritten I'm afraid. The HA architecture requires exposing air quality as a separate sensor and that's currently not possible. (You could add it as a custom component though.)

Mariusthvdb:
Yes, I'd love these to be available in Ha. 
Can we write a rest sensor for that maybe. Must investigate.

ggravlingen:
@Mariusthvdb if you mean in HA, the current implementation observes the state of a light, or of a fan. The fan, however, has certain properties such as fan speed or air quality. The fan speed is exposed as a "fan" device. We would also have to setup a separate sensor device for the air quality. Multiple observation of the same device (the fan) is currently not possible.

My understanding is that we should implement this for this to work:
https://developers.home-assistant.io/docs/integration_fetching_data/#coordinated-single-api-poll-for-data-for-all-entities

But this is a separate topic from this PR so let's just discuss the proposed changes here.

pytradfri/device/air_purifier.py
MartinHjelmare: I'm asking you if you think this. Is your answer that it never will be interesting to know the mode?

@Mariusthvdb please don't divert the conversation to another topic. If you want to ask a question, do that in a separate conversation.

ggravlingen: No, in the way the device currently work I find the suggested changes to be the most appropriate. I cannot see why we should to consider the fan speed as being modes.

MartinHjelmare: Ok.

MartinHjelmare (approved):

PR #406, opened 2022-01-15
Create example_jsonlist.py

add30417:
The standard output of the script, is a list with the devices + status information in JSON Format.

### DIFF
--- a/examples/example_jsonlist.py
+++ b/examples/example_jsonlist.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python3
+"""
+This is an example of how the pytradfri-library can be used async to create a list in json format (e.g. for a webserver, to retrieve information about TRADFRI system).
+
+To run the script, do the following:
+$ pip3 install pytradfri
+$ Download this file (example_jsonlist.py)
+$ python3 example_jsonlist.py) <IP>
+
+Where <IP> is the address to your IKEA gateway. The first time
+running you will be asked to input the 'Security Code' found on
+the back of your IKEA gateway.
+"""
+
+import os
+
+# Hack to allow relative import above top level package
+import sys
+
+folder = os.path.dirname(os.path.abspath(__file__))
+sys.path.insert(0, os.path.normpath("%s/.." % folder))
+
+import argparse
+import asyncio
+import uuid
+
+from pytradfri import Gateway
+from pytradfri.api.aiocoap_api import APIFactory
+from pytradfri.error import PytradfriError
+from pytradfri.util import load_json, save_json
+
+#CONFIG_FILE = "tradfri_standalone_psk.conf"
+
+parser = argparse.ArgumentParser()
+parser.add_argument(
+        "host", metavar="IP", type=str, help="IP Address of your Tradfri gateway"
+        )
+parser.add_argument(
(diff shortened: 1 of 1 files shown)


examples/example_jsonlist.py
```diff
@@ -0,0 +1,114 @@
...
+
+    gateway = Gateway()
+
+    devices_command = gateway.get_devices()
+    devices_commands = await api(devices_command)
+    devices = await api(devices_commands)
+
+    list_devices = [dev for dev in devices]
+    #print("{\"DEVICES\":[", end= '');
+    print("[", end= '');
```
ggravlingen, line 92: Any way we can turn this into a util function and take it as an argument in the existing example file?

ggravlingen (commented):
Haven't seen anyone request this feature before. I'm curious, what is the use case?

examples/example_jsonlist.py
```diff
@@ -0,0 +1,114 @@
...
+import argparse
+import asyncio
+import uuid
+
+from pytradfri import Gateway
+from pytradfri.api.aiocoap_api import APIFactory
+from pytradfri.error import PytradfriError
+from pytradfri.util import load_json, save_json
+
+#CONFIG_FILE = "tradfri_standalone_psk.conf"
```
ggravlingen, line 32: We should remove stale comments.

examples/example_jsonlist.py
```diff
@@ -0,0 +1,114 @@
...
+    api = api_factory.request
+
+    gateway = Gateway()
+
+    devices_command = gateway.get_devices()
+    devices_commands = await api(devices_command)
+    devices = await api(devices_commands)
+
+    list_devices = [dev for dev in devices]
+    #print("{\"DEVICES\":[", end= '');
```
ggravlingen, line 91: We should remove stale comments.

examples/example_jsonlist.py
```diff
@@ -0,0 +1,114 @@
...
+            print("\",\"TYPE\":\"SOCKET\"", end='');
+        elif device.has_light_control:
+            print(device.light_control.lights[0].dimmer, end='');
+            print("\",\"TYPE\":\"LIGHT\"", end='');
+        else:
+            print("\",\"TYPE\":\"UNKOWN\"", end='');
+        print(",\"TYPENAME\":\"" + device.device_info.model_number + "\",\"REACHABLE\":\"" + str(device.reachable) + "\"}", end='');
+        first=False;
+    print("]");
+    #print("]}");
```
ggravlingen, line 109: We should remove stale comments.

add30417:
I’m using this feature, to retrieve the current status for my smart home system.

examples/example_jsonlist.py
add30417: Stollen? It is kind of fork from other example. Why not use good comments?

ggravlingen: Not stolen, comments that are commented out, ie # before them

add30417: Sorry, well of course can be removed.

examples/example_jsonlist.py
add30417: Well, it is depending of design guideline. I would keep it separated. Also better for future change and extensions at both files.

PR #411, opened 2022-01-29
Remove vscode settings from git

MartinHjelmare:
-  We don't want to track vscode settings but allow each contributor to have their own settings.


ggravlingen (approved):

ggravlingen:
Thanks!

PR #412, opened 2022-01-29
Fix observation err_callback typing

MartinHjelmare:
- The type annotation for the observation error callback was incorrect. Fix that.
- I needed to fix and some additional things around the observation method in the APIs to correct typing after the error callback fix.


ggravlingen (approved):
Sweet!

PR #413, opened 2022-01-29
Add typing for DeviceInfo class

ggravlingen:
The `DeviceInfo` class is not typed. This PR fixes that.


pytradfri/device/__init__.py
```diff
@@ -136,48 +137,51 @@ class DeviceInfo:
     }
     ATTR_BATTERY = "9"
 
-    def __init__(self, device):
+    def __init__(self, device) -> None:
```
MartinHjelmare, line 140:
```suggestion
    def __init__(self, device: Device) -> None:
```

pytradfri/device/__init__.py
```diff
@@ -136,48 +137,51 @@ class DeviceInfo:
...
         """Create object of class."""
         self._device = device
 
     @property
-    def manufacturer(self):
+    def manufacturer(self) -> str:
         """Human readable manufacturer name."""
-        return self.raw.get(DeviceInfo.ATTR_MANUFACTURER)
+        return cast(str, self.raw.get(DeviceInfo.ATTR_MANUFACTURER))
```
MartinHjelmare, line 147: We can't cast to string and use dict.get at the same time. `dict.get` may return None.

pytradfri/device/__init__.py
```diff
@@ -136,48 +137,51 @@ class DeviceInfo:
...
     @property
-    def battery_level(self):
+    def battery_level(self) -> int:
         """Battery in 0..100."""
-        return self.raw.get(DeviceInfo.ATTR_BATTERY)
+        return cast(int, self.raw.get(DeviceInfo.ATTR_BATTERY))
 
     @property
-    def raw(self):
+    def raw(self) -> dict:
```
MartinHjelmare, line 185: We could return a TypedDict here to make more clear what data is returned.

Long term we should use pydantic to type the returned api responses. That will simplify the typing a lot.

MartinHjelmare (changes requested):

pytradfri/device/__init__.py
ggravlingen: What is the best-practice way to go about here?
```
try:
    return cast(str, self.raw[DeviceInfo.ATTR_MANUFACTURER])
except KeyError:
    return None
```
or
```
if DeviceInfo.ATTR_MANUFACTURER in self.raw:
    return cast(str, self.raw[DeviceInfo.ATTR_MANUFACTURER])

return None
```

Or maybe another option that I'm unaware of?

MartinHjelmare: I think we should either use a TypedDict or pydantic to type the raw response so that we don't need to cast the type of the raw items in the properties here. I also believe that some of the device info raw items are not optional.

ggravlingen: Ok, I’ll look into that and see how it works.

pytradfri/device/__init__.py
```diff
@@ -136,48 +138,49 @@ class DeviceInfo:
...
+    def __init__(self, device: Device) -> None:
         """Create object of class."""
         self._device = device
 
     @property
-    def manufacturer(self):
+    def manufacturer(self) -> str:
         """Human readable manufacturer name."""
-        return self.raw.get(DeviceInfo.ATTR_MANUFACTURER)
+        return self.raw[DeviceInfo.ATTR_MANUFACTURER]  # type: ignore
```
ggravlingen, line 148: It seems using variables together with `TypedDict` isn't allowed:
`pytradfri/device/__init__.py:153: error: TypedDict key must be a string literal; expected one of ("0", "1", "2", "3", "6", ...)  [misc]`

What I found when searching on this issue was the recommendation to ignore the typing error. Perhaps we should drop the use of variables and simply use `return self.raw["0"]` instead? I've added documentation of what the keys contain to the typing file.

pytradfri/types.py
```diff
@@ -0,0 +1,17 @@
+"""Typing information."""
+from typing import TypedDict
+
+TypeDeviceInfo = TypedDict(
```
MartinHjelmare, line 4: Will we need this type in more places than in the device module? Otherwise I suggest we keep the definition there to make it easier to read and follow the code.

pytradfri/device/__init__.py
MartinHjelmare: I think we can drop the constants if we only use them in this class.

pytradfri/device/__init__.py
```diff
@@ -134,50 +131,53 @@ class DeviceInfo:
...
         """Represent current power source."""
-        if DeviceInfo.ATTR_POWER_SOURCE not in self.raw:
+        if "6" not in self.raw:
             return None
+
         return DeviceInfo.VALUE_POWER_SOURCES.get(self.power_source, "Unknown")
 
     @property
-    def battery_level(self):
+    def battery_level(self) -> Optional[int]:
```
ggravlingen, line 173: Judging by the unit tests, this is optional today. Don't want to change that here.

pytradfri/types.py
ggravlingen: An idea I had was to create a `LightType´ and a ´AirPurifierType´, which would include the ´TypeDeviceInfo´. But that's for a future PR so I'll move it.

ggravlingen:
Hmm, tests for Python 3.7 fail and I now see that `TypedDict` is Python 3.8+, which is probably the culprit. I propose dropping support for Python 3.7 (in a separate PR) and then rebasing this branch on top of that.

MartinHjelmare:
Yes, we can drop Python 3.7. 👍

MartinHjelmare (approved):
Nice!

PR #414, opened 2022-01-30
Drop support for Python 3.7, cleanup Dockerfiles

ggravlingen:
This PR drops support for Python 3.7.

Also edited four of the Dockerfiles:
- Dockerfile: update to Python 3.10.
- DockerfileRPI: switched to another base image. I can't test this on my RPI as it runs Home Assistant OS. Given what I read [here](https://www.balena.io/docs/reference/base-images/base-images/), this image should work on a RPI.
- DockerfileRPIW: deleted, don't see the point of having separate images for separate versions of RPI.
- DockerfileRPIW-installed: deleted, don't see the point of having separate images for separate versions of RPI.


tox.ini
```diff
@@ -1,10 +1,9 @@
 [tox]
-envlist = py37, py38, py39, py310, lint, typing, coverage
+envlist = py38, py39, py310, lint, typing, coverage
 skip_missing_interpreters = True
 
 [gh-actions]
 python =
-  3.7: py37, lint, typing, coverage
   3.8: py38
```
MartinHjelmare, line 7:
```suggestion
  3.8: py38, lint, typing, coverage
```

MartinHjelmare (approved):

PR #419, opened 2022-02-05
Add attributes for the AirPurifier

ggravlingen:
This PR adds attributes that are available on the AirPurifier fan. Some information regarding the properties have been gathered from this post: https://github.com/dresden-elektronik/deconz-rest-plugin/issues/5351#issuecomment-968267581


pytradfri/const.py
```diff
@@ -21,6 +21,11 @@
 ATTR_BLIND_CURRENT_POSITION = "5536"
 ATTR_BLIND_TRIGGER = "5523"
 
+ATTR_AIR_PURIFIER_FILTER_RUNTIME: Final = "5902"
```
MartinHjelmare, line 24: Maybe sort the constants according to their number starting with "5900"?

pytradfri/device/air_purifier.py
```diff
@@ -55,6 +60,31 @@ def raw(self) -> TypeAirPurifier:
             cast(TypeRawList, self.device.raw)[ROOT_AIR_PURIFIER][self.index],
         )
 
+    @property
+    def motor_runtime_total(self) -> int:
+        """Return runtime of fan motor, expressed in minutes."""
+        return self.raw[ATTR_AIR_PURIFIER_MOTOR_RUNTIME_TOTAL]
+
+    @property
+    def filter_lifetime_total(self) -> int:
```
MartinHjelmare, line 69: I'd sort the properties 🔡.

pytradfri/device/air_purifier.py
```diff
@@ -55,6 +60,31 @@ def raw(self) -> TypeAirPurifier:
...
+        """Return total lifetime of filter, expressed in minutes."""
+        return self.raw[ATTR_AIR_PURIFIER_FILTER_LIFETIME_TOTAL]
+
+    @property
+    def filter_status(self) -> int:
+        """Return filter status."""
```
MartinHjelmare, line 75: What does this integer represent in terms of filter status?

ggravlingen: I am not sure to be honest as I haven't had my filter for more than three months (its lifetime seems to be six months). If I were to make a guesss it will return `1` when its time to replace the filter. But I'm not willing to put that into the documentation of the property yet.

MartinHjelmare: Does it return 0 today for you?

MartinHjelmare (approved):

pytradfri/device/air_purifier.py
ggravlingen: Yes:

```
{'5900': 1,
 '5902': 152245,
 '5903': 0,
 '5904': 259200,
 '5905': 0,
 '5906': 1,
 '5907': 8,
 '5908': 10,
 '5909': 152244,
 '5910': 106955,
 '9003': 0}
```

PR #420, opened 2022-02-05
Bump version to 9.0.0

ggravlingen:


pytradfri/VERSION
```diff
@@ -1 +1 @@
-8.0.1
+8.1.0
```
MartinHjelmare, line 1: Technically, removing support for Python 3.7 is a breaking change. Should we bump major version?

ggravlingen: Let’s be prudent, I’ll bump major.

MartinHjelmare (approved):

PR #426, opened 2022-02-12
Add type annotation for Blind class

ggravlingen:
This PR implements a `TypedDict` for the `Blind` class.

### DIFF
--- a/pytradfri/device/blind.py
+++ b/pytradfri/device/blind.py
@@ -1,14 +1,25 @@
 """Represent a blind."""
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, cast
+from typing import TYPE_CHECKING, TypedDict, cast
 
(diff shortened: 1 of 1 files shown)


ggravlingen:
In the next PR, I will move the typedicts to a new file and start chaining them together.

MartinHjelmare:
I you're planning to do bigger changes, I recommend start working with pydantic instead. That should be the end goal.

ggravlingen:
> 

I've been doing some reading and can't really tell what's the big advantage? Or is it just a fact that it's considered best-practice for describing data?

I'm easily convinced so I don't mind figuring out how pydantic works. Regardless, what I'd like to do is merge this and then the next PR where I've chained together a few TypedDicts. That increases my understanding a bit before I move to another library.

MartinHjelmare (approved):

PR #427, opened 2022-02-12
Type hints for ApiResource

ggravlingen:
Add type `TypedDict` for `ApiResource`, which combines previously added TypedDicts.

Following this PR, we will refactor and start using `pydantic`.

### DIFF
--- a/pytradfri/typing.py
+++ b/pytradfri/typing.py
@@ -6,20 +6,6 @@ as keys: https://www.python.org/dev/peps/pep-0589/#alternative-syntax
 """
 from typing import List, TypedDict
 
-TypeDeviceInfo = TypedDict(
-    "TypeDeviceInfo",
-    {
-        "0": str,  # Manufacturer
-        "1": str,  # Model number
-        "2": str,  # Serial number
-        "3": str,  # Firmware version
-        "6": int,  # Power source
-        "7": str,  # OTA image type
-        "9": int,  # Battery level
-    },
-)
-
-
 TypeAirPurifier = TypedDict(
     "TypeAirPurifier",
     {
@@ -42,6 +28,19 @@ TypeBlind = TypedDict(
     {
         "5536": int,  # Current blind position
         "9003": int,  # ID
+    },
+)
+
+TypeDeviceInfo = TypedDict(
+    "TypeDeviceInfo",
+    {
+        "0": str,  # Manufacturer
+        "1": str,  # Model number
+        "2": str,  # Serial number
+        "3": str,  # Firmware version
+        "6": int,  # Power source
+        "7": str,  # OTA image type
+        "9": int,  # Battery level
     },
 )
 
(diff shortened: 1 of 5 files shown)


pytradfri/typing.py
```diff
@@ -0,0 +1,59 @@
...
+TypeBlind = TypedDict(
+    "TypeBlind",
+    {
+        "5536": int,  # Current blind position
+        "9003": int,  # ID
+    },
+)
+
+
+TypeApiResource = TypedDict(
```
ggravlingen, line 49: Maybe `APIResponse is a more appropriate` name?

MartinHjelmare: Yeah, let's try to mimic the names we already have for each thing and add `Response` at the end. It will make even more sense when using Pydantic models, I think.

MartinHjelmare: So I'd make it `APIResourceResponse` if we don't want to change the existing name.

ggravlingen: `ApiResourceResponse` it is. Also change the other ones.

ggravlingen: Allright, fixed now. My git-hook seem to have been disabled.

pytradfri/typing.py
```diff
@@ -0,0 +1,58 @@
...
+        "9": int,  # Battery level
+    },
+)
+
+
+ApiResourceResponse = TypedDict(
+    "ApiResourceResponse",
+    {
+        "9001": str,  # Name
+        "9002": int,  # Created at
```
MartinHjelmare, line 52: The `Created at` item may not be in the response according to our existing code. Should we allow this with `total=False`?

https://docs.python.org/3/library/typing.html#typing.TypedDict

pytradfri/typing.py
```diff
@@ -0,0 +1,58 @@
...
+        "9003": str,  # ID
+        "3": DeviceInfoResponse,
+        "15025": List[AirPurifierResponse],
```
MartinHjelmare, line 55: These are also not always present. It will get easier to describe this relationship later with Pydantic I think, so we don't have to think too much about it now though.

pytradfri/typing.py
ggravlingen: Hmm, it seems to be present in all example devices so I'm not sure why that is the case in the code now:
https://github.com/home-assistant-libs/pytradfri/blob/master/tests/devices.py

But maybe we should refrain from changing how the code behaves and just add `total=False`?

pytradfri/typing.py
ggravlingen: Ok, I'll leave it as is for now.

pytradfri/typing.py
MartinHjelmare: Ok, let's leave it like this for now and make a mental note to investigate this more. Maybe the existing code is just wrong and this item is always present in the response?

MartinHjelmare (approved):

PR #428, opened 2022-02-12
Add pydantic to requirements

ggravlingen:
This PR adds `pydantic` to the list of requirements so what we can add type hints using the [library](https://pydantic-docs.helpmanual.io/).

Also adding suggested plugin settings for `mypy`:
https://pydantic-docs.helpmanual.io/mypy_plugin/#enabling-the-plugin

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 aiocoap==0.4.3
 DTLSSocket==0.1.12
-
+pydantic==1.9.0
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -1,3 +1,3 @@
 aiocoap==0.4.3
 DTLSSocket==0.1.12
-
+pydantic==1.9.0
```
MartinHjelmare, line 3: We should add it to `install_requires` in setup.py too, without the version limit.

MartinHjelmare (approved):

PR #429, opened 2022-02-13
Implement pydantic

ggravlingen:
This PR implements `pydantic` for type hints.


ggravlingen:
@MartinHjelmare this got a bit bigger than I had expected for this PR, as it seem I have to add models for all devices in order to make this work. I most likely also have to update a few of the tests as they look for values using dict keys. I'll keep this as a draft until I get all tests to pass locally. No need to review the full extent of this code until then.

I do, hoever, have a few questions that you possibly know the answer to. Will address them as comments in the code.

pytradfri/resource.py
```diff
@@ -18,26 +18,27 @@
 class ApiResource:
     """Base object for resources returned from the gateway."""
 
-    def __init__(self, raw: ApiResourceResponse) -> None:
+    def __init__(self, raw: TypeRawList) -> None:
         """Initialize base object."""
-        self.raw = raw
+        self.raw: ApiResourceResponse = ApiResourceResponse(**raw)
```
ggravlingen, line 23: My understanding of the library is that this is how we populate the pydantic-model with data?

pytradfri/type_hint.py
```diff
@@ -0,0 +1,105 @@
...
+    state: int = Field(alias=ATTR_DEVICE_STATE)
+    dimmer: int = Field(alias=ATTR_LIGHT_DIMMER)
+
+
+class ApiResourceResponse(BaseModel):
+    """Represent an API response."""
+
+    air_purifier: list[AirPurifierResponse] | None = Field(alias=ROOT_AIR_PURIFIER)
+    application_type: int = Field(alias=ATTR_APPLICATION_TYPE)
+    blind: list[BlindResponse] | None = Field(alias=ATTR_START_BLINDS)
```
ggravlingen, line 93: How would we handle the case where there is no data for blinds, ie for a light device. I've set it to `None` here, but it leads to other complications, like here where it tries to iterate over None.

https://github.com/home-assistant-libs/pytradfri/blob/c9daaa16d88b33c7ba20b2298e3a16be5b5eca41/pytradfri/device/blind.py#L25

I'd prefer not having to check for `None` everywhere when representing various states of a device. There surely is a better way?

ggravlingen: Maybe overwrite one of the methods of `BaseModel` so that it only creates the field if the relevant key is present in the `dict` we pass to it?

pytradfri/type_hint.py
```diff
@@ -0,0 +1,105 @@
...
+    ATTR_OTA_UPDATE_STATE,
+    ATTR_REACHABLE_STATE,
+    ATTR_START_BLINDS,
+    ATTR_SWITCH_PLUG,
+    ROOT_AIR_PURIFIER,
+    ROOT_SIGNAL_REPEATER,
+)
+
+
+class AirPurifierResponse(BaseModel):
```
MartinHjelmare, line 44: I'd move each response model to the module where its used. Eg move `AirPurifierResponse` to the `air_purifier.py` module.

pytradfri/type_hint.py
MartinHjelmare: We may be able to improve this, but I don't think it's a big problem right now. We'll only need to assert the item to satisfy typing once in the device module, eg blind.py, and once in the device controller module, eg blind_controller.py.

pytradfri/resource.py
MartinHjelmare: Example:
```py
class ApiResource:
    """Base object for resources returned from the gateway."""

    _model_class: type[ResourceResponse]  = ResourceResponse
    raw: ResourceResponse

    def __init__(self, raw: TypeRaw) -> None:
        """Initialize base object."""
        self.raw = self._model_class(**raw)


class Device(ApiResource):
    """Base class for devices."""

    _model_class: type[DeviceResponse] = DeviceResponse
    raw: DeviceResponse
```

pytradfri/type_hint.py
ggravlingen: I moved them to one file as I was having problems with circular imports. That shouldn't be an issue with your proposal of inheriting from the resource class. I'll move them back.

pytradfri/resource.py
ggravlingen: Thank you for the pointers! I'll explore this further and see what I can come up with.

ggravlingen:
I've made some changes but no need to check here yet. I'm trying to find my way around this.

pytradfri/resource.py
MartinHjelmare: If we don't want to update all modules in this PR we can add some conditional logic around the `raw` instantiation in the base resource class and set the `_model_class` as optional and keep the old way if `_model_class` is None.

ggravlingen:
I ended up adding responses for all device types. Figure it might be easier than adding workarounds to handle different `raw` return types.

Tests pass locally now, although I notice they are failing here. Had to rewrite a few of them to fit into the new ways of accessing data.

Also having issues with mypy that I can't figure out at the moment:
```
pytradfri/device/socket.py:33: error: Value of type "Optional[List[SocketResponse]]" is not indexable  [index]
pytradfri/device/signal_repeater.py:28: error: Value of type "Optional[List[SignalRepeaterResponse]]" is not indexable  [index]
pytradfri/device/blind.py:32: error: Value of type "Optional[List[BlindResponse]]" is not indexable  [index]
pytradfri/device/air_purifier.py:56: error: Value of type "Optional[List[AirPurifierResponse]]" is not indexable  [index]
pytradfri/device/socket_control.py:19: error: Incompatible return value type (got "Optional[List[SocketResponse]]", expected "List[SocketResponse]")  [return-value]
pytradfri/device/light.py:58: error: Value of type "Optional[List[LightResponse]]" is not indexable  [index]
pytradfri/device/blind_control.py:23: error: Incompatible return value type (got "Optional[List[BlindResponse]]", expected "List[BlindResponse]")  [return-value]
pytradfri/device/air_purifier_control.py:25: error: Incompatible return value type (got "Optional[List[AirPurifierResponse]]", expected "List[AirPurifierResponse]")  [return-value]
pytradfri/device/light_control.py:73: error: Incompatible return value type (got "Optional[List[LightResponse]]", expected "List[LightResponse]")  [return-value]
pytradfri/device/__init__.py:136: error: Item "None" of "Optional[DeviceInfoResponse]" has no attribute "manufacturer"  [union-attr]
pytradfri/device/__init__.py:141: error: Item "None" of "Optional[DeviceInfoResponse]" has no attribute "model_number"  [union-attr]
pytradfri/device/__init__.py:146: error: Item "None" of "Optional[DeviceInfoResponse]" has no attribute "serial"  [union-attr]
pytradfri/device/__init__.py:151: error: Item "None" of "Optional[DeviceInfoResponse]" has no attribute "firmware_version"  [union-attr]
pytradfri/device/__init__.py:156: error: Item "None" of "Optional[DeviceInfoResponse]" has no attribute "power_source"  [union-attr]
pytradfri/device/__init__.py:161: error: Item "None" of "Optional[DeviceInfoResponse]" has no attribute "power_source"  [union-attr]
pytradfri/device/__init__.py:162: error: Item "None" of "Optional[DeviceInfoResponse]" has no attribute "power_source"  [union-attr]
pytradfri/device/__init__.py:169: error: Item "None" of "Optional[DeviceInfoResponse]" has no attribute "battery_level"  [union-attr]
pytradfri/device/__init__.py:170: error: Item "None" of "Optional[DeviceInfoResponse]" has no attribute "battery_level"  [union-attr]
``` 

The way I see it, all device types must be optional as not all responses contain data for all device types. But setting a device as optional leads to errors like this: `Optional[List[SocketResponse]]" is not indexable  [index]`

The same goes for `Optional[DeviceInfoResponse]`. I had to set `DeviceInfo` to optional as smart task responses don't have that information.

pytradfri/device/air_purifier.py
```diff
@@ -15,22 +17,43 @@
...
+    id: int = Field(alias=ATTR_ID)
+    leds_off: int = Field(alias=ATTR_AIR_PURIFIER_LEDS_OFF)
+    mode: int = Field(alias=ATTR_AIR_PURIFIER_MODE)
+    motor_runtime_total: int = Field(alias=ATTR_AIR_PURIFIER_MOTOR_RUNTIME_TOTAL)
+
+
 class AirPurifier:
     """Represent an air purifier."""
 
+    _model_class: type[AirPurifierResponse] = AirPurifierResponse
```
MartinHjelmare, line 49: I think you misunderstood me. The different device types don't need a model class since they don't inherit from ApiResource. It's `Device`, `Group`, `SmartTask` etc that need a specific model class to differentiate their response from the `ApiResourceResponse` of their parent.

pytradfri/device/air_purifier.py
```diff
@@ -15,22 +17,43 @@
...
+    raw: AirPurifierResponse
+
     def __init__(self, device: Device, index: int) -> None:
         """Create object of class."""
         self.device = device
         self.index = index
+        self.raw = self.device.raw.air_purifier[index]
```
MartinHjelmare, line 56: We can't set `raw` as an instance attribute since the `device.raw` attribute is replaced during an update.

pytradfri/device/air_purifier.py
```diff
@@ -84,24 +107,19 @@ def is_auto_mode(self) -> bool:
...
-        return self.raw[ATTR_AIR_PURIFIER_LEDS_OFF] == 1
+        return self.raw.leds_off == 1
 
     @property
     def motor_runtime_total(self) -> int:
         """Return runtime of fan motor, expressed in minutes."""
-        return self.raw[ATTR_AIR_PURIFIER_MOTOR_RUNTIME_TOTAL]
-
-    @property
-    def raw(self) -> AirPurifierResponse:
```
MartinHjelmare, line 100: We need to keep this property.

pytradfri/resource.py
```diff
@@ -5,39 +5,102 @@
...
+from .device.light import LightResponse
+from .device.socket import SocketResponse
 
 # type alias
 TypeRaw = Dict[str, Union[str, int, List[Dict[str, Union[str, int]]]]]
 TypeRawSimple = Dict[str, Union[str, int]]
 TypeRawList = Dict[str, List[Dict[str, Union[str, int]]]]
 
 
+class DeviceInfoResponse(BaseModel):
```
MartinHjelmare, line 45: We want to move this to the device.py module but we can do that in a follow up PR to avoid having to move around the names inside device.py.

pytradfri/smart_task.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, gateway, raw):
         """Initialize the class."""
         super().__init__(raw)
         self._gateway = gateway
+        self.task_raw: TypeRawSimple = raw
```
MartinHjelmare, line 56: Instead of changing here, I suggest adding checks in ApiResource and keep the old code inside cases for the resources we don't want to update now (mood, group and smart task).

pytradfri/resource.py
```diff
@@ -5,39 +5,102 @@
...
+        alias=ROOT_SIGNAL_REPEATER
+    )
+
+
 class ApiResource:
     """Base object for resources returned from the gateway."""
 
-    def __init__(self, raw: ApiResourceResponse) -> None:
+    _model_class: type[ApiResourceResponse] = ApiResourceResponse
+    raw: ApiResourceResponse
```
MartinHjelmare, line 81: We can do it like this to allow us to wait with updating all of the resource children.
```suggestion
    _model_class: type[ResourceResponse] | None = None
    raw: TypeRaw | ResourceResponse
```

pytradfri/resource.py
```diff
@@ -5,39 +5,102 @@
...
+
+    def __init__(self, raw: TypeRawList) -> None:
         """Initialize base object."""
-        self.raw = raw
+        self.raw = self._model_class(**raw)
```
MartinHjelmare, line 85:
```suggestion
        if self._model_class:
            self.raw = self._model_class(**raw)
        else:
            self.raw = raw
```

pytradfri/resource.py
```diff
@@ -5,39 +5,102 @@
...
 
     @property
-    def id(self) -> str | None:
+    def id(self) -> int | None:
         """Id."""
-        return self.raw[ATTR_ID]
+        return self.raw.id
```
MartinHjelmare, line 90:
```suggestion
        if self._model_class:
            resource_id = self.raw.id  # type: ignore[union-attr]
        else:
            resource_id = self.raw[ATTR_ID]  # type: ignore[index]
        return resource_id
```

pytradfri/resource.py
```diff
@@ -5,39 +5,102 @@
...
 
     @property
     def name(self) -> str | None:
         """Name."""
-        return self.raw[ATTR_NAME]
+        return self.raw.name
```
MartinHjelmare, line 95:
```suggestion
        if self._model_class:
            name = self.raw.name  # type: ignore[union-attr]
        else:
            name = self.raw[ATTR_NAME]  # type: ignore[index]
        return name
```

pytradfri/resource.py
```diff
@@ -5,39 +5,102 @@
...
     @property
     def created_at(self) -> datetime | None:
         """Return timestamp of creation."""
-        if ATTR_CREATED_AT not in self.raw:
-            return None
-        return datetime.utcfromtimestamp(self.raw[ATTR_CREATED_AT])
+        if self.raw.created_at:
+            return datetime.utcfromtimestamp(self.raw.created_at)
+
+        return None
```
MartinHjelmare, line 103:
```suggestion
        if self._model_class:
            created_at = self.raw.created_at  # type: ignore[union-attr]
        else:
            created_at = self.raw[ATTR_CREATED_AT]  # type: ignore[index]

        if created_at is None:
            return None
        return datetime.utcfromtimestamp(created_at)
```

pytradfri/resource.py
```diff
@@ -5,39 +5,102 @@
...
 
     @property
     @abstractmethod
```
MartinHjelmare, line 106: Update the `observe_callback` and `process_result` below in a similar way to keep compatibility with the other resources that we don't update now.

pytradfri/device/air_purifier.py
ggravlingen: Looking at your proposed changes, I certainly did. I will go through and try to implement.

ggravlingen:
Want me to remove all type hints I've added and make this PR a bit cleaner? It's not a biggie making a new PR and adding them back again later.

MartinHjelmare:
> Want me to remove all type hints I've added and make this PR a bit cleaner? It's not a biggie making a new PR and adding them back again later.

If it's touching files that we're changing anyway here, it's ok to keep. But we should to keep the number of files changed down. 👍

pytradfri/smart_task.py
ggravlingen: Not sure I understood what you wanted me to not do here?

tox.ini
```diff
@@ -27,12 +27,14 @@ deps =
 
 [testenv:typing]
 deps =
+  -rrequirements.txt
```
ggravlingen, line 30: Looks like we need `pydantic` here, and we add it in `requiremenst.txt`

ggravlingen:
We have a test that works when I run it as a single file locally but not when I run the entire test suite. Other than that one test, mypy now succeeds.

pytradfri/device/__init__.py
```diff
@@ -14,38 +19,56 @@
 class Device(ApiResource):
     """Base class for devices."""
 
+    _model_class = ResourceResponse
+
     @property
     def application_type(self) -> int | None:
         """Return application type."""
-        return self.raw.application_type
+        if self._model_class:
```
MartinHjelmare, line 27: We know that model class is not None here. We don't need any cases here.

pytradfri/resource.py
```diff
@@ -53,7 +53,7 @@ class DeviceInfoResponse(BaseModel):
     battery_level: int | None = Field(alias=ATTR_DEVICE_BATTERY)
 
 
-class ApiResourceResponse(BaseModel):
+class ResourceResponse(BaseModel):
```
MartinHjelmare, line 56: We want a base ApiResourceResponse that doesn't have any device specific attributes. Then make a separate response model for the Device class that has the additional device attributes.

pytradfri/resource.py
```diff
@@ -65,11 +65,13 @@ class ApiResourceResponse(BaseModel):
     ota_update_state: int | None = Field(alias=ATTR_OTA_UPDATE_STATE)
     reachable: int | None = Field(alias=ATTR_REACHABLE_STATE)
 
-    air_purifier: list[AirPurifierResponse] | None = Field(alias=ROOT_AIR_PURIFIER)
-    blind: list[BlindResponse] | None = Field(alias=ATTR_START_BLINDS)
-    light: list[LightResponse] | None = Field(alias=ATTR_LIGHT_CONTROL)
-    socket: list[SocketResponse] | None = Field(alias=ATTR_SWITCH_PLUG)
-    signal_repater: list[SignalRepeaterResponse] | None = Field(
+    air_purifier_control: list[AirPurifierResponse] | None = Field(
```
MartinHjelmare, line 68: These attributes should go under a `DeviceResponse` model.

pytradfri/device/__init__.py
```diff
@@ -14,38 +19,56 @@
 class Device(ApiResource):
     """Base class for devices."""
 
+    _model_class = ResourceResponse
```
MartinHjelmare, line 22:
```suggestion
    _model_class = DeviceResponse
```

pytradfri/smart_task.py
MartinHjelmare: Please revert these changes.

pytradfri/device/signal_repeater.py
```diff
@@ -1,16 +1,28 @@
 """Represent a signal repeater."""
-from ..const import ROOT_SIGNAL_REPEATER
+from __future__ import annotations
```
MartinHjelmare, line 2: Please revert the changes to all files that we don't need to do change otherwise.

tests/test_color.py
```diff
@@ -8,6 +8,7 @@
     SUPPORT_RGB_COLOR,
     SUPPORT_XY_COLOR,
 )
+from pytradfri.device.light import LightResponse
```
MartinHjelmare, line 11: We shouldn't need to change any tests in this PR.

pytradfri/device/air_purifier.py
```diff
@@ -15,15 +17,32 @@
...
+    ATTR_ID,
 )
-from ..typing import AirPurifierResponse
 
 if TYPE_CHECKING:
     # avoid cyclic import at runtime.
     from . import Device
 
 
+class AirPurifierResponse(BaseModel):
```
MartinHjelmare, line 28: Don't add any new models besides the DeviceResponse model now. The aim should just be to update the device module and keep other changes to a minimum.

pytradfri/resource.py
MartinHjelmare: Here's how I would write the `DeviceResponse` model:

```py
class DeviceResponse(ApiResourceResponse):
    """Represent a device response."""

    # Type with Any for now to allow smaller typing work chunks

    air_purifier_control: Optional[List[AirPurifierResponse]] = Field(
        alias=ROOT_AIR_PURIFIER
    )
    application_type: int = Field(alias=ATTR_APPLICATION_TYPE)
    blind_control: Optional[List[BlindResponse]] = Field(alias=ATTR_START_BLINDS)
    device_info: DeviceInfoResponse = Field(alias=ATTR_DEVICE_INFO)
    last_seen: Optional[int] = Field(alias=ATTR_LAST_SEEN)
    light_control: Optional[List[Dict[str, Any]]] = Field(alias=ATTR_LIGHT_CONTROL)
    reachable_state: int = Field(alias=ATTR_REACHABLE_STATE)
    signal_repeater_control: Optional[List[Dict[str, Any]]] = Field(
        alias=ROOT_SIGNAL_REPEATER
    )
    socket_control: Optional[List[Dict[str, Any]]] = Field(alias=ATTR_SWITCH_PLUG)
```

pytradfri/resource.py
MartinHjelmare: Example:

```py
class ApiResourceResponse(BaseModel):
    """Represent a resource response."""

    id: int = Field(alias=ATTR_ID)
    name: str = Field(alias=ATTR_NAME)
    created_at: Optional[int] = Field(alias=ATTR_CREATED_AT)
```

ggravlingen: Thank you!

pytradfri/resource.py
ggravlingen: Thank you!

MartinHjelmare:
We'll need to update the different device type modules `raw` property like so:

```py
#  air_purifier.py
@property
def raw(self) -> AirPurifierResponse:
    """Return raw data that it represents."""
    air_purifier_control_response = self.device.raw.air_purifier_control
    assert air_purifier_control_response is not None
    return air_purifier_control_response[self.index]
```

```py
#  air_purifier_control.py
@property
def raw(self) -> list[AirPurifierResponse]:
    """Return raw data that it represents."""
    air_purifier_control_response = self._device.raw.air_purifier_control
    assert air_purifier_control_response is not None
    return air_purifier_control_response
```

pytradfri/device/signal_repeater.py
ggravlingen: I have rolled back all changes now

ggravlingen:
> We'll need to update the different device type modules `raw` property like so:
> 
> ```python
> #  air_purifier.py
> @property
> def raw(self) -> AirPurifierResponse:
>     """Return raw data that it represents."""
>     air_purifier_control_response = self.device.raw.air_purifier_control
>     assert air_purifier_control_response is not None
>     return air_purifier_control_response[self.index]
> ```
> 
> ```python
> #  air_purifier_control.py
> @property
> def raw(self) -> list[AirPurifierResponse]:
>     """Return raw data that it represents."""
>     air_purifier_control_response = self._device.raw.air_purifier_control
>     assert air_purifier_control_response is not None
>     return air_purifier_control_response
> ```

I'm curious, what does the `assert` statement do here? Does it override/disable some sort of internal check that allows us to proceed?

MartinHjelmare:
The response attributes of the sub types on the `DeviceResponse` model are optional. But we know that when we've created a device sub type, eg air purifier, we know that that part of the response is no longer optional. We use the assert statement to inform mypy that it's not optional anymore. It's similar to using `cast` but less intrusive as we keep the rest of the type information intact.

ggravlingen:
> 


> The response attributes of the sub types on the `DeviceResponse` model are optional. But we know that when we've created a device sub type, eg air purifier, we know that that part of the response is no longer optional. We use the assert statement to inform mypy that it's not optional anymore. It's similar to using `cast` but less intrusive as we keep the rest of the type information intact.

Thank you and thank you for guiding me, again. There is absolutely no way I would have figured that out myself.

pytradfri/device/__init__.py
```diff
@@ -33,93 +15,110 @@ class Device(ApiResource):
     """Base class for devices."""
 
     @property
-    def application_type(self):
+    def _model_class(self) -> type[DeviceResponse]:  # type: ignore[override]
+        """Return model class type."""
+        return DeviceResponse
```
MartinHjelmare, line 20:
```suggestion
    _model_class: type[DeviceResponse] = DeviceResponse
    raw: DeviceResponse
```

pytradfri/device/__init__.py
```diff
@@ -14,7 +14,10 @@
 class Device(ApiResource):
     """Base class for devices."""
 
-    _model_class: type[DeviceResponse] = DeviceResponse
+    @property
+    def _model_class(self) -> type[DeviceResponse]:  # type: ignore[override]
```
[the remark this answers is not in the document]
ggravlingen: I received some weird error but let me try your solution.

pytradfri/device/__init__.py
```diff
@@ -33,93 +15,109 @@ class Device(ApiResource):
     """Base class for devices."""
 
     @property
```
MartinHjelmare, line 17:
```suggestion
```

pytradfri/resource.py
```diff
@@ -3,41 +3,130 @@
...
+    created_at: int = Field(alias=ATTR_CREATED_AT)
+    id: int = Field(alias=ATTR_ID)
+    name: str | None = Field(alias=ATTR_NAME)
+
+
 class ApiResource:
     """Base object for resources returned from the gateway."""
 
-    def __init__(self, raw: ApiResourceResponse) -> None:
+    _model_class: type[ResourceResponse] | None = None
```
MartinHjelmare, line 91:
```suggestion
    _model_class: type[ApiResourceResponse] | None = None
```

pytradfri/resource.py
```diff
@@ -3,41 +3,130 @@
...
+
+
+class ResourceResponse(BaseModel):
+    """Represent an API response."""
+
+    created_at: int = Field(alias=ATTR_CREATED_AT)
+    id: int = Field(alias=ATTR_ID)
+    name: str | None = Field(alias=ATTR_NAME)
+
+
```
MartinHjelmare, line 87:
```suggestion
```

pytradfri/resource.py
```diff
@@ -3,41 +3,130 @@
...
 class ApiResource:
     """Base object for resources returned from the gateway."""
 
-    def __init__(self, raw: ApiResourceResponse) -> None:
+    _model_class: type[ResourceResponse] | None = None
+    raw: TypeRaw | ResourceResponse
```
MartinHjelmare, line 92:
```suggestion
    raw: TypeRaw | ApiResourceResponse
```

pytradfri/resource.py
```diff
@@ -3,41 +3,130 @@
...
 TypeRawList = Dict[str, List[Dict[str, Union[str, int]]]]
 
 
+class ApiResourceResponse(BaseModel):
+    """Represent a resource response."""
+
+    id: int = Field(alias=ATTR_ID)
+    name: str = Field(alias=ATTR_NAME)
+    created_at: Optional[int] = Field(alias=ATTR_CREATED_AT)
+    ota_update_state: int | None = Field(alias=ATTR_OTA_UPDATE_STATE)
```
MartinHjelmare, line 46: We can't use the `|` syntax to mark things optional in pydantic. We need to use `Optional`.

pytradfri/device/__init__.py
```diff
@@ -2,26 +2,8 @@
 from datetime import datetime
```
MartinHjelmare, line 2:
```suggestion
from __future__ import annotations

from datetime import datetime
```

pytradfri/device/signal_repeater_control.py
```diff
@@ -1,5 +1,6 @@
 """Class to control the signal repeater."""
-from ..const import ROOT_SIGNAL_REPEATER
+from typing import Any, Dict
```
MartinHjelmare, line 2:
```suggestion
from __future__ import annotations

from typing import Any
```

pytradfri/device/signal_repeater_control.py
```diff
@@ -8,9 +9,11 @@ class SignalRepeaterControl(BaseController):
     """Class to control the signal repeaters."""
 
     @property
-    def raw(self):
+    def raw(self) -> list[Dict[str, Any]]:
```
MartinHjelmare, line 12:
```suggestion
    def raw(self) -> list[dict[str, Any]]:
```

pytradfri/device/signal_repeater_control.py
```diff
@@ -8,9 +9,11 @@ class SignalRepeaterControl(BaseController):
...
         """Return raw data that it represents."""
-        return self._device.raw[ROOT_SIGNAL_REPEATER]
+        signal_repater_control_response = self._device.raw.signal_repeater_control
+        assert signal_repater_control_response is not None
+        return signal_repater_control_response
```
MartinHjelmare, line 16:
```suggestion
        signal_repeater_control_response = self._device.raw.signal_repeater_control
        assert signal_repeater_control_response is not None
        return signal_repeater_control_response
```

pytradfri/device/signal_repeater.py
```diff
@@ -1,5 +1,5 @@
 """Represent a signal repeater."""
-from ..const import ROOT_SIGNAL_REPEATER
+from typing import Any, Dict
```
MartinHjelmare, line 2:
```suggestion
from __future__ import annotations

from typing import Any
```

pytradfri/device/signal_repeater.py
```diff
@@ -11,6 +11,8 @@ def __init__(self, device, index):
         self.index = index
 
     @property
-    def raw(self):
+    def raw(self) -> Dict[str, Any]:
```
MartinHjelmare, line 14:
```suggestion
    def raw(self) -> dict[str, Any]:
```

pytradfri/device/signal_repeater.py
```diff
@@ -11,6 +11,8 @@ def __init__(self, device, index):
...
         """Return raw data that it represents."""
-        return self.device.raw[ROOT_SIGNAL_REPEATER][self.index]
+        signal_repater_control_response = self.device.raw.signal_repeater_control
+        assert signal_repater_control_response is not None
+        return signal_repater_control_response[self.index]
```
MartinHjelmare, line 18:
```suggestion
        signal_repeater_control_response = self.device.raw.signal_repeater_control
        assert signal_repeater_control_response is not None
        return signal_repeater_control_response[self.index]
```

pytradfri/device/socket.py
```diff
@@ -1,5 +1,7 @@
 """Represent a socket."""
-from ..const import ATTR_DEVICE_STATE, ATTR_SWITCH_PLUG
+from typing import Any, Dict
```
MartinHjelmare, line 2:
```suggestion
from __future__ import annotations

from typing import Any
```

pytradfri/device/socket.py
```diff
@@ -16,9 +18,11 @@ def state(self):
         return self.raw.get(ATTR_DEVICE_STATE) == 1
 
     @property
-    def raw(self):
+    def raw(self) -> Dict[str, Any]:
```
MartinHjelmare, line 21:
```suggestion
    def raw(self) -> dict[str, Any]:
```

pytradfri/device/socket_control.py
```diff
@@ -1,4 +1,6 @@
 """Class to control the sockets."""
+from typing import Any, Dict
```
MartinHjelmare, line 2:
```suggestion
from __future__ import annotations

from typing import Any
```

pytradfri/device/socket_control.py
```diff
@@ -9,9 +11,11 @@ class SocketControl(BaseController):
     """Class to control the sockets."""
 
     @property
-    def raw(self):
+    def raw(self) -> list[Dict[str, Any]]:
```
MartinHjelmare, line 14:
```suggestion
    def raw(self) -> list[dict[str, Any]]:
```

pytradfri/resource.py
```diff
@@ -3,41 +3,130 @@
...
+
+class DeviceInfoResponse(BaseModel):
+    """Represent the device info part of the device response."""
+
+    manufacturer: str = Field(alias=ATTR_DEVICE_MANUFACTURER)
+    model_number: str = Field(alias=ATTR_DEVICE_MODEL_NUMBER)
+    serial: str = Field(alias=ATTR_DEVICE_SERIAL)
+    firmware_version: str = Field(alias=ATTR_DEVICE_FIRMWARE_VERSION)
+    power_source: int | None = Field(alias=ATTR_DEVICE_POWER_SOURCE)
+    battery_level: int | None = Field(alias=ATTR_DEVICE_BATTERY)
```
MartinHjelmare, line 57:
```suggestion
    power_source: Optional[int] = Field(alias=ATTR_DEVICE_POWER_SOURCE)
    battery_level: Optional[int] = Field(alias=ATTR_DEVICE_BATTERY)
```

MartinHjelmare:
We also need to modify the imports at the top of `typing.py`:
```py
import sys
from typing import List

if sys.version_info < (3, 9, 2):
    from typing_extensions import TypedDict
else:
    from typing import TypedDict
```

MartinHjelmare:
With those changes, mypy and pytest pass for me locally.

pytradfri/resource.py
MartinHjelmare: Note that `ota_update_state` is a new attribute.

ggravlingen:
> With those changes, mypy and pytest pass for me locally.

Phew! Again, thank you!

ggravlingen:
We could also clean up `typing.py` but I'll submit another PR doing that when this is merged.

MartinHjelmare (approved):
Awesome!

PR #432, opened 2022-02-26
Move response types to their correct location

ggravlingen:


pytradfri/device/air_purifier_control.py
```diff
@@ -10,10 +12,33 @@
...
-from ..typing import AirPurifierResponse
 from .air_purifier import AirPurifier
 from .base_controller import BaseController
 
+if sys.version_info < (3, 9, 2):
+    from typing_extensions import TypedDict
+else:
+    from typing import TypedDict
+
+AirPurifierResponse = TypedDict(
```
MartinHjelmare, line 23: If we define it in `air_purifier.py` our import flow we be cleaner. We'll need to guard less imports under `if TYPE_CHECKING`.

PR #435, opened 2022-02-26
Implement pydantic for sockets

ggravlingen:


pytradfri/device/socket.py
```diff
@@ -1,32 +1,46 @@
 """Represent a socket."""
```
MartinHjelmare, line 1: We should add the socket modules to mypy.ini strict typing.

pytradfri/device/socket_control.py
```diff
@@ -1,34 +1,32 @@
...
         return socket_control_response
 
     @property
-    def sockets(self):
+    def sockets(self) -> list[Socket]:
         """Return socket objects of the socket control."""
         return [Socket(self._device, i) for i in range(len(self.raw))]
 
-    def set_state(self, state, *, index=0):
+    def set_state(self, state: int, *, index=0) -> Command:
```
MartinHjelmare, line 25:
```suggestion
    def set_state(self, state: int, *, index: int = 0) -> Command:
```

pytradfri/device/socket_control.py
```diff
@@ -1,34 +1,32 @@
...
         """Set state of a socket."""
-        return self.set_values({ATTR_DEVICE_STATE: int(state)}, index=index)
+        return self.set_values({ATTR_DEVICE_STATE: state}, index=index)
 
-    def set_values(self, values, *, index=0):
+    def set_values(self, values: dict[str, int], *, index) -> Command:
```
MartinHjelmare, line 29: Should we keep the default index? Otherwise it's a breaking change.

pytradfri/device/socket.py
MartinHjelmare: We're also missing this for the air purifier.

pytradfri/device/socket_control.py
ggravlingen: Yes, that was an unintended removal.

ggravlingen:
I ended up having to change an unrelated file. Should I revert and add that file to `mypy.ini` in a separate PR?

MartinHjelmare (approved):

MartinHjelmare:
You decide. It's ok I think.

ggravlingen:
> You decide. It's ok I think.

Then I leave it and move on to lights.

PR #436, opened 2022-02-26
Implement pydantic for lights

ggravlingen:


ggravlingen:
Figured we'd sort out the typing stuff before I move on to fixing the tests.

pytradfri/device/light.py
```diff
@@ -27,69 +48,70 @@ class Light:
...
+    def state(self) -> bool:
         """Return device state."""
-        return self.raw.get(ATTR_DEVICE_STATE) == 1
+        return self.raw.state == 1
 
     @property
-    def dimmer(self):
+    def dimmer(self) -> int | None:
         """Return dimmer if present."""
         if self.supported_features & SUPPORT_BRIGHTNESS:
```
MartinHjelmare, line 69: dimmer is not optional in the response model. Can it still be not supported?

pytradfri/device/light_control.py
```diff
@@ -110,9 +118,11 @@ def set_color_temp(self, color_temp, *, index=0, transition_time=None):
 
         return self.set_values(values, index=index)
 
-    def set_hex_color(self, color, *, index=0, transition_time=None):
+    def set_hex_color(
+        self, color: str, *, index: int = 0, transition_time: int | None = None
+    ) -> Command:
         """Set hex color of the light."""
-        values = {
+        values: dict[str, Any] = {
```
MartinHjelmare, line 125:
```suggestion
        values: dict[str, str | int] = {
```

pytradfri/device/light_control.py
```diff
@@ -161,7 +186,7 @@ def set_predefined_color(self, colorname, *, index=0, transition_time=None):
         except KeyError:
             raise ColorError(f"Invalid color specified: {colorname}") from KeyError
 
-    def set_values(self, values, *, index=0):
+    def set_values(self, values: dict[str, Any], *, index: int = 0) -> Command:
```
MartinHjelmare, line 189:
```suggestion
    def set_values(self, values: dict[str, str | int], *, index: int = 0) -> Command:
```

pytradfri/device/light.py
```diff
@@ -27,69 +48,70 @@ class Light:
...
+    def hex_color(self) -> str | None:
         """Return hex color."""
         if self.supported_features & SUPPORT_HEX_COLOR:
-            return self.raw.get(ATTR_LIGHT_COLOR_HEX)
+            return self.raw.color_hex
         return None
 
     @property
-    def xy_color(self):
+    def xy_color(self) -> tuple[int | None, int | None] | None:
```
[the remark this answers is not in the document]
ggravlingen: I agree, didn't want to make too many changes in the first iteration.

pytradfri/device/light.py
ggravlingen: I can't think of any light device that is not dimmable. Perhaps we should drop this supported_features-check and just return the int?

MartinHjelmare: Yeah, maybe. Ok to leave for the future though.

pytradfri/device/light_control.py
```diff
@@ -163,7 +166,10 @@ def set_hsb(
         self._value_validate(hue, RANGE_HUE, "Hue")
         self._value_validate(saturation, RANGE_SATURATION, "Saturation")
 
-        values = {ATTR_LIGHT_COLOR_SATURATION: saturation, ATTR_LIGHT_COLOR_HUE: hue}
+        values: dict[str, str | int] = {
```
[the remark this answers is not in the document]
ggravlingen: I received a slew of errors if I didn't. But that could of course be due to some other error. I'll remove it and post the error I get here.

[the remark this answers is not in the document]
ggravlingen:
```
pytradfri/device/light_control.py:181: error: Argument 1 to "set_values" of "LightControl" has incompatible type "Dict[str, int]"; expected "Dict[str, Union[str, int]]"  [arg-type]
pytradfri/device/light_control.py:181: note: "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
pytradfri/device/light_control.py:181: note: Consider using "Mapping" instead, which is covariant in the value type
Found 1 error in 1 file (checked 27 source files)
```

pytradfri/device/light.py
```diff
@@ -27,69 +48,74 @@ class Light:
...
+            self.supported_features & SUPPORT_XY_COLOR
+            and self.raw.color_xy_x is not None
+            and self.raw.color_xy_y is not None
+        ):
+            return (self.raw.color_xy_x, self.raw.color_xy_y)
         return None
 
     @property
-    def hsb_xy_color(self):
+    def hsb_xy_color(
```
ggravlingen, line 99: Maybe add checks for `None` here as well?

pytradfri/device/light_control.py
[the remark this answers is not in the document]
MartinHjelmare: I pushed a fix.

pytradfri/device/light.py
ggravlingen: https://github.com/home-assistant/core/blob/1ee4f47f8be3cd56f00e8fec35ce445436c80b82/homeassistant/components/tradfri/light.py#L200

MartinHjelmare: Side note for the future: Since we return xy color from another property I would expect this one to only return hsb color. Maybe add a property that does that and deprecate this one.

I agree that we should only return a tuple if all items are not None.

ggravlingen: Want me to deprecate here in this PR?

MartinHjelmare: Nah, just some thoughts for the future.

tests/test_device.py
```diff
@@ -352,12 +352,6 @@ def test_socket_state_on():
     assert socket.state is True
 
 
-def test_set_state_none(device):
```
ggravlingen, line 355: Should we keep this? We can't send `None` to set_state as it's typed for `int` now.

tests/test_color.py
```diff
@@ -29,30 +30,36 @@
...
         + SUPPORT_XY_COLOR
     )
 
     assert (
-        supported_features(LIGHT_CWS["3311"][0])
+        supported_features(LightResponse(**LIGHT_CWS["3311"][0]))
         == SUPPORT_BRIGHTNESS + SUPPORT_RGB_COLOR + SUPPORT_HEX_COLOR + SUPPORT_XY_COLOR
     )
 
+    x = supported_features(LightResponse(**LIGHT_CWS_CUSTOM_COLOR["3311"][0]))
```
MartinHjelmare, line 59:
```suggestion
```

tests/test_color.py
```diff
@@ -29,30 +30,36 @@
...
+    print(x)
```
MartinHjelmare, line 60:
```suggestion
```

tests/test_device.py
```diff
@@ -425,30 +418,39 @@ def test_has_light_control_false():
 
 
 # Test light state function
-def test_light_state_on(device):
+def test_light_state_on():
     """Test light on."""
+    device_data = deepcopy(LIGHT_WS)
+    device_data[ATTR_LIGHT_CONTROL][0][ATTR_DEVICE_STATE] = 1
+    device = Device(deepcopy(device_data))
```
MartinHjelmare, line 425: Why do we copy twice?

tests/test_device.py
MartinHjelmare: I don't think we need this test. Although during run time we could still pass None, code wise it would fail our static typing checks.

ggravlingen: OK, then I'll leave it deleted

ggravlingen:
When this is merged, we should be good to bump in `core` I guess?

MartinHjelmare (approved):
Nice!

MartinHjelmare:
I would complete the typing so we can add py.typed before bumping in core if it's only for typing purposes.

PR #442, opened 2022-03-04
Type hint gateway

ggravlingen:


ggravlingen:
I believe this constant might need some consideration when we've sorted out type hints on the Gateway class:
https://github.com/home-assistant-libs/pytradfri/blob/a979ed1a0d3a985509400466776d5628db21dc85/pytradfri/command.py#L7

ggravlingen:
Maybe it would make more sense to add type hints to `SmartTask` and `Mood` before doing this? 🤔

pytradfri/gateway.py
```diff
@@ -67,103 +92,103 @@ def process_result(result):
...
-    def get_group(self, group_id):
+    def get_group(self, group_id: str) -> Command:
         """
         Return specified group.
 
         Returns a Command.
         """
 
-        def process_result(result):
+        def process_result(result: TypeRaw):
```
MartinHjelmare, line 139:
```suggestion
        def process_result(result: TypeRaw) -> Group:
```

pytradfri/gateway.py
```diff
@@ -50,14 +54,35 @@ def process_result(result):
         )
 
     @classmethod
-    def get_endpoints(cls):
+    def get_endpoints(cls) -> Command:
         """
         Return all available endpoints on the gateway.
 
+        The response from the gateway looks like this:
+        <//15006>;
```
MartinHjelmare, line 62: The example here will better match how we split the string of we make each line end with the comma.

MartinHjelmare:
> I believe this constant might need some consideration when we've sorted out type hints on the Gateway class:
> 
> https://github.com/home-assistant-libs/pytradfri/blob/a979ed1a0d3a985509400466776d5628db21dc85/pytradfri/command.py#L7

We can change it to return `Optional[Any]` for now. I have some ideas how we can improve that later but that's for when we want to improve typing in the command.py module.

ggravlingen:
> > I believe this constant might need some consideration when we've sorted out type hints on the Gateway class:
> > https://github.com/home-assistant-libs/pytradfri/blob/a979ed1a0d3a985509400466776d5628db21dc85/pytradfri/command.py#L7
> 
> We can change it to return `Optional[Any]` for now. I have some ideas how we can improve that later but that's for when we want to improve typing in the command.py module.

I ended up doing the same with the `data` argument.

ggravlingen:
Would you oppose if we deprecated moods and smart tasks? I for one don't use them and don't really feel they're worth maintaining. For users who need those features, we could point to Home Assistant.

MartinHjelmare:
I think the library should represent the tradfri api as close as possible. I don't think we should remove it.

pytradfri/command.py
```diff
@@ -14,12 +14,12 @@ def __init__(
...
         method: str,
         path: list[str],
-        data: dict[str, Any] | None = None,
+        data: Any | None = None,
         *,
         parse_json: bool = True,
         observe: bool = False,
         observe_duration: int = 0,
-        process_result: TypeProcessResultCb = None,
+        process_result: Optional[Any] = None,
```
MartinHjelmare, line 22: We should update `TypeProcessResultCb`. We still know it's a callback. It's just the return value we want to be more lax about.

MartinHjelmare (commented):
Are you planning to type the rest of the gateway module in this PR?

pytradfri/gateway.py
```diff
@@ -67,140 +91,142 @@ def process_result(result):
...
 
         Returns a Command.
         """
 
-        def process_result(result):
+        def process_result(result: TypeRaw) -> Mood:
             return Mood(result, mood_parent)
 
+        # The path argument of Command expects a list of strings
+        assert isinstance(mood_parent, str)
```
MartinHjelmare, line 194: If `mood_parent` needs to be a string we should fix the signature above and not make it optional. It's a bug fix.

pytradfri/command.py
ggravlingen: > I think the library should represent the tradfri api as close as possible. I don't think we should remove it.

OK, then it'll remain.

ggravlingen:
> Are you planning to type the rest of the gateway module in this PR?

No, the rest of the file will be a follow-up PR. Felt that was a bit easier to comprehend.

pytradfri/gateway.py
```diff
@@ -67,140 +91,142 @@ def process_result(result):
...
 
-        def process_result(result):
+        def process_result(result: list[str]) -> list[Command]:
             return [self.get_mood(mood, mood_parent=group_id) for mood in result]
 
         return Command("get", [ROOT_MOODS, group_id], process_result=process_result)
 
     @classmethod
-    def get_mood(cls, mood_id, *, mood_parent=None):
+    def get_mood(cls, mood_id: str, *, mood_parent: str | None = None) -> Command:
```
MartinHjelmare, line 183:
```suggestion
    def get_mood(cls, mood_id: str, *, mood_parent: str) -> Command:
```

pytradfri/gateway.py
MartinHjelmare, line 194:
```suggestion
```

ggravlingen:
I hadn't planned on touching `command.py` as it looks to be strictly typed already. Do you want to add your ideas to the file yourself?

MartinHjelmare:
Yeah, I'll look at that separately. :+1:

MartinHjelmare (approved):
Nice!

PR #443, opened 2022-03-05
Fix light and socket set_state

MartinHjelmare:
- We changed the signature of these methods when introducing typing. I suggest we keep the boolean type for state in the signature to avoid a breaking change, and to keep the type more narrow.


ggravlingen (approved):
I agree!

PR #444, opened 2022-03-05
Type aiocoap api

MartinHjelmare:
- See https://github.com/home-assistant-libs/pytradfri/pull/388 for the previous attempt.
- I've corrected the type for the `Command.result` as that depends on the return value of the `process_result` callback.
- I'm planning to continue working on the command typing to make it more useful in follow up PRs.

### DIFF
--- a/pytradfri/api/aiocoap_api.py
+++ b/pytradfri/api/aiocoap_api.py
@@ -2,9 +2,11 @@
 from __future__ import annotations
 
 import asyncio
+from collections.abc import Callable
+from enum import Enum
 import json
 import logging
-from typing import Callable
+from typing import cast
 
 from aiocoap import Context, Message
 from aiocoap.credentials import CredentialsMissingError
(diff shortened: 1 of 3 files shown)


pytradfri/api/aiocoap_api.py
```diff
@@ -177,33 +198,22 @@ async def _execute(self, api_command, timeout):
 
         return api_command.result
 
-    def debug_comm(self, call_type, api_commands):
```
ggravlingen, line 180: This was only used to figure out the various types of returns we could get from the gateway I presume?

ggravlingen: Oh, I see now that you refactored it..

pytradfri/api/aiocoap_api.py
```diff
@@ -15,56 +17,73 @@
...
-_SENTINEL = object()
+
+
+class UndefinedType(Enum):
+    """Singleton type for use with not set sentinel values."""
+
+    _singleton = 0
+
+
+_SENTINEL = UndefinedType._singleton  # pylint: disable=protected-access
```
ggravlingen, line 35: I have never seen this type of code before. What does it do?

MartinHjelmare: It's to type the singleton that is used to mark empty value.

https://www.python.org/dev/peps/pep-0484/#support-for-singleton-types-in-unions

ggravlingen: Thanks!

ggravlingen (approved):
Sweet!

PR #445, opened 2022-03-05
Add type hints for GatewayInfo

ggravlingen:
This PR:
- Implements pydantic for the `GatewayInfo` response

We can't add the file to the strict typing list yet as we need to add type hints for `group.py` and `mood.py` first.

### DIFF
--- a/pytradfri/gateway.py
+++ b/pytradfri/gateway.py
@@ -373,7 +373,7 @@ class GatewayInfo:
 
     @property
     def time_source(self) -> int:
-        """Return homekit id."""
+        """Return time source."""
         return self.raw.time_source
 
     @property
--- a/pytradfri/gateway.py
+++ b/pytradfri/gateway.py
@@ -320,7 +320,7 @@ class GatewayInfo:
 
     @property
     def firmware_version(self) -> str:
-        """NTP server in use."""
+        """Return gateway firmware version."""
         return self.raw.firmware_version
 
     @property
--- a/pytradfri/gateway.py
+++ b/pytradfri/gateway.py
@@ -2,11 +2,15 @@
 from __future__ import annotations
 
 from datetime import datetime
-from typing import Any
+from typing import Any, Optional
+
+from pydantic import BaseModel, Field
 
 from .command import Command
 from .const import (
+    ATTR_ALEXA_PAIR_STATUS,
     ATTR_AUTH,
+    ATTR_CERTIFICATE_PROV,
     ATTR_COMMISSIONING_MODE,
     ATTR_CURRENT_TIME_ISO8601,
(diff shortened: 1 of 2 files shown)


pytradfri/gateway.py
```diff
@@ -259,74 +288,119 @@ def factory_reset(cls) -> Command:
 class GatewayInfo:
     """This class contains Gateway information."""
 
-    def __init__(self, raw):
+    _model_class: type[GatewayInfoResponse] = GatewayInfoResponse
```
MartinHjelmare, line 291: We don't need this attribute since we don't inherit from ApiResponse. We can just create the `raw` instance directly from `GatewayInfoResponse` in the init method below.
```suggestion
```

pytradfri/gateway.py
```diff
@@ -259,74 +288,119 @@ def factory_reset(cls) -> Command:
...
+    raw: GatewayInfoResponse
+
+    def __init__(self, raw: TypeRaw):
         """Create object of class."""
-        self.raw = raw
+        self.raw = self._model_class(**raw)
```
MartinHjelmare, line 296:
```suggestion
        self.raw = GatewayInfoResponse(**raw)
```

pytradfri/gateway.py
```diff
@@ -259,74 +288,119 @@ def factory_reset(cls) -> Command:
...
-    def update(self):
+    def update(self) -> Command:
         """
         Update the info.
 
         Returns a Command.
         """
 
-        def process_result(result):
+        def process_result(result: GatewayInfoResponse) -> None:
```
MartinHjelmare, line 398:
```suggestion
        def process_result(result: TypeRaw) -> None:
```

pytradfri/gateway.py
```diff
@@ -259,74 +288,119 @@ def factory_reset(cls) -> Command:
...
             """Define callback to process result."""
             self.raw = result
```
MartinHjelmare, line 400:
```suggestion
            self.raw = GatewayInfoResponse(**result)
```

pytradfri/gateway.py
```diff
@@ -259,74 +288,119 @@ def factory_reset(cls) -> Command:
 class GatewayInfo:
     """This class contains Gateway information."""
 
-    def __init__(self, raw):
+    _model_class: type[GatewayInfoResponse] = GatewayInfoResponse
+    raw: GatewayInfoResponse
+
+    def __init__(self, raw: TypeRaw):
```
MartinHjelmare, line 294:
```suggestion
    def __init__(self, raw: TypeRaw) -> None:
```

MartinHjelmare (approved):

PR #446, opened 2022-03-05
Add type hints for Mood class

ggravlingen:


pytradfri/mood.py
```diff
@@ -1,22 +1,25 @@
...
         """Create object of class."""
         super().__init__(raw)
         self._parent = parent
 
     @property
-    def path(self):
+    def path(self) -> list[str]:
         """Path."""
-        return [ROOT_MOODS, self._parent, self.id]
+        return [ROOT_MOODS, self._parent, cast(str, self.id)]
```
MartinHjelmare, line 21: The `id` is typed as an integer. So we need to copy to string if we want a string. We can only use `cast` when we want to tell mypy what type something is. It doesn't change the type.
```suggestion
        return [ROOT_MOODS, self._parent, str(self.id)]
```

ggravlingen: OK, thanks!

MartinHjelmare (approved):

ggravlingen:
I'll fix the test

ggravlingen:
I fixed it but should we really have to convert to str in the path property? We should have receved a string from parent class?

ggravlingen:
Will not merge until you have commented on the last change.

pytradfri/mood.py
```diff
@@ -1,22 +1,23 @@
...
         """Create object of class."""
         super().__init__(raw)
         self._parent = parent
 
     @property
-    def path(self):
+    def path(self) -> list[str]:
         """Path."""
-        return [ROOT_MOODS, self._parent, self.id]
+        return [ROOT_MOODS, str(self._parent), str(self.id)]
```
MartinHjelmare, line 19: Yeah, the parent should be a string already according to our typing. We need to check that if we think it's not true.

ggravlingen: Odd, it looks like strings here:
https://github.com/home-assistant-libs/pytradfri/blob/2fc8fa0ec1a639a0ace6093b1311d2009d453123/pytradfri/gateway.py#L212

ggravlingen: Oh, I believe moods linked to a group is of type `int` and not `str`. Let me push that code and we can have a look at both at the same time.

MartinHjelmare: We didn't resolve this, I think.

ggravlingen: Sorry, I misunderstood you in the other PR. This is what I meant by converting to str and the looking into it later. Should we revert PR?

PR #447, opened 2022-03-05
Add type hints for Group class

ggravlingen:


ggravlingen:
Notes for myself for tomorrow, check if ID is always an int. If so, we should make necessary changes everywhere.

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
 )
 from .error import ColorError
-from .resource import ApiResource
+from .resource import ApiResource, TypeRaw
+
+if TYPE_CHECKING:
+    from .gateway import Gateway
+
+
+class GroupResponse(BaseModel):
```
MartinHjelmare, line 41: Doesn't the group response have the ApiResourceResponse as base?

ggravlingen: It does, should it inherit? It does make sense now that you mention it - they are sharing some properties. I'll refactor.

ggravlingen:
> Notes for myself for tomorrow, check if ID is always an int. If so, we should make necessary changes everywhere.

`id` (key 9003) is an int in all our examples, I wonder why `path` is typed as a list of strings?

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
+    from .gateway import Gateway
+
+
+class GroupResponse(ApiResourceResponse):
+    """Represent API response for a blind."""
+
+    color_hex: Optional[str] = Field(alias=ATTR_LIGHT_COLOR_HEX)
+    dimmer: int = Field(alias=ATTR_LIGHT_DIMMER)
+    group_members: Optional[dict[str, Any]] = Field(alias=ATTR_GROUP_MEMBERS)
+    id: int = Field(alias=ATTR_ID)
```
MartinHjelmare, line 47: We only need to add the attributes that the parent doesn't provide.
```suggestion
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
 
 
 class Group(ApiResource):
     """Represent a group."""
 
-    def __init__(self, gateway, raw):
+    _model_class: type[GroupResponse] = GroupResponse
+    raw: GroupResponse
+
+    def __init__(self, gateway: Gateway, raw: TypeRaw):
```
MartinHjelmare, line 58:
```suggestion
    def __init__(self, gateway: Gateway, raw: TypeRaw) -> None:
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
         """Create object of class."""
         super().__init__(raw)
+        self.raw = GroupResponse(**raw)
```
MartinHjelmare, line 61: This is handled by the parent.
```suggestion
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
+        if self.raw.color_hex:
+            return self.raw.color_hex
+
+        return None
 
     @property
-    def member_ids(self):
-        """Members of this group."""
-        info = self.raw.get(ATTR_GROUP_MEMBERS, {})
+    def member_ids(self) -> list | list[int]:
```
MartinHjelmare, line 88: We don't need a special type case for returning an empty list.
```suggestion
    def member_ids(self) -> list[int]:
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
+if TYPE_CHECKING:
+    from .gateway import Gateway
+
+
+class GroupResponse(ApiResourceResponse):
+    """Represent API response for a blind."""
+
+    color_hex: Optional[str] = Field(alias=ATTR_LIGHT_COLOR_HEX)
+    dimmer: int = Field(alias=ATTR_LIGHT_DIMMER)
+    group_members: Optional[dict[str, Any]] = Field(alias=ATTR_GROUP_MEMBERS)
```
MartinHjelmare, line 46:
```suggestion
    group_members: Optional[Dict[str, Dict[str, List[int]]]] = Field(alias=ATTR_GROUP_MEMBERS)
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
+        info = self.raw.group_members
 
-        if not info or ATTR_HS_LINK not in info:
+        if info is None or ATTR_HS_LINK not in info:
             return []
 
-        return info[ATTR_HS_LINK].get(ATTR_ID, [])
+        member_ids: list[int] = info[ATTR_HS_LINK][ATTR_ID]
+
+        if member_ids:
```
MartinHjelmare, line 102: I don't think we need this check unless this can be None and we need to change our typing of the response.

pytradfri/resource.py
```diff
@@ -26,8 +29,8 @@ class ApiResourceResponse(BaseModel):
 class ApiResource:
     """Base object for resources returned from the gateway."""
 
-    _model_class: type[ApiResourceResponse] | None = None
-    raw: TypeRaw | ApiResourceResponse
+    _model_class: type[ApiResourceResponse] | type[GroupResponse] | None = None
```
MartinHjelmare, line 32: We don't need this change.

tests/test_group.py
```diff
@@ -63,7 +63,7 @@ def test_setters(group):
     assert cmd.data == {ATTR_GROUP_ID: GROUP[ATTR_ID], ATTR_ID: [65547]}
 
 
-def test_moods(group):
+def test_moods(group: Group):
```
MartinHjelmare, line 66:
```suggestion
def test_moods(group: Group) -> None:
```

pytradfri/group.py
ggravlingen: I'm not sure to be honest but it would make sense that it returns an empty list rather than nothing.

ggravlingen:
Should we just convert the ints to string to fit the current implementation of `path` and then address this later?

MartinHjelmare:
Yeah, that's fine.

pytradfri/group.py
MartinHjelmare: Yes, but according to our current typing. It could be an empty list from the beginning but then we don't need this check.

Do we know that this can be None?

ggravlingen:
> Yeah, that's fine.

OK, I'll fix and this here and merge the other PR where that has already been done.

pytradfri/group.py
ggravlingen: I don't know if it can be None.

ggravlingen: I wasn't intending to change the property's behaviour, only rewrite it for the pydantic class.

MartinHjelmare: We have changed it already since we're now using `dict[key]` instead of `dict.get` and to access the `ATTR_ID` key. So we have changed the meaning of the optionality, from `ATTR_ID` may be missing to `ATTR_ID` could hold a `None` value.

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
 
     @property
-    def mood_id(self):
+    def mood_id(self) -> str:
         """Active mood."""
-        return self.raw.get(ATTR_MOOD)
+        return self.raw.mood_id
 
-    def members(self):
+    def members(self) -> list[Command]:
```
MartinHjelmare, line 112:
```suggestion
    def members(self) -> list[Command[Device]]:
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
         """Return device objects of members of this group."""
-        return [self._gateway.get_device(dev) for dev in self.member_ids]
+        return [self._gateway.get_device(str(dev)) for dev in self.member_ids]
 
-    def add_member(self, memberid):
+    def add_member(self, memberid: str) -> Command:
```
MartinHjelmare, line 116:
```suggestion
    def add_member(self, memberid: str) -> Command[None]:
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
         """Add a member to this group."""
         return self._gateway.add_group_member(
             {ATTR_GROUP_ID: self.id, ATTR_ID: [memberid]}
         )
 
-    def remove_member(self, memberid):
+    def remove_member(self, memberid: str) -> Command:
```
MartinHjelmare, line 122:
```suggestion
    def remove_member(self, memberid: str) -> Command[None]:
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
         """Remove a member from this group."""
         return self._gateway.remove_group_member(
             {ATTR_GROUP_ID: self.id, ATTR_ID: [memberid]}
         )
 
-    def moods(self):
+    def moods(self) -> Command:
```
MartinHjelmare, line 128:
```suggestion
    def moods(self) -> Command[list[Command[Mood]]]:
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
         """Return mood objects of moods in this group."""
-        return self._gateway.get_moods(self.id)
+        return self._gateway.get_moods(str(self.id))
 
-    def mood(self):
+    def mood(self) -> Command:
```
MartinHjelmare, line 132:
```suggestion
    def mood(self) -> Command[Mood]:
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
         """Remove a member from this group."""
         return self._gateway.remove_group_member(
             {ATTR_GROUP_ID: self.id, ATTR_ID: [memberid]}
         )
 
-    def moods(self):
+    def moods(self) -> Command:
         """Return mood objects of moods in this group."""
-        return self._gateway.get_moods(self.id)
+        return self._gateway.get_moods(str(self.id))
```
MartinHjelmare, line 130:
```suggestion
        return self._gateway.get_moods(self.id)
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
 
-    def mood(self):
+    def mood(self) -> Command:
         """Active mood."""
-        return self._gateway.get_mood(self.mood_id, mood_parent=self.id)
+        return self._gateway.get_mood(self.mood_id, mood_parent=str(self.id))
```
MartinHjelmare, line 134:
```suggestion
        return self._gateway.get_mood(self.mood_id, mood_parent=self.id)
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
 
-    def activate_mood(self, mood_id):
+    def activate_mood(self, mood_id: str) -> Command:
```
MartinHjelmare, line 136:
```suggestion
    def activate_mood(self, mood_id: str) -> Command[None]:
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
         """Activate a mood."""
         return self.set_values({ATTR_MOOD: mood_id, ATTR_DEVICE_STATE: int(self.state)})
 
-    def set_state(self, state):
+    def set_state(self, state: bool) -> Command:
```
MartinHjelmare, line 140:
```suggestion
    def set_state(self, state: bool) -> Command[None]:
```

pytradfri/group.py
```diff
@@ -24,85 +32,116 @@
...
         """Set state of a group."""
         return self.set_values({ATTR_DEVICE_STATE: int(state)})
 
-    def set_dimmer(self, dimmer, transition_time=None):
+    def set_dimmer(self, dimmer: int, transition_time: int | None = None) -> Command:
```
MartinHjelmare, line 144:
```suggestion
    def set_dimmer(
        self, dimmer: int, transition_time: int | None = None
    ) -> Command[None]:
```

pytradfri/group.py
```diff
@@ -115,7 +154,9 @@ def set_dimmer(self, dimmer, transition_time=None):
             values[ATTR_TRANSITION_TIME] = transition_time
         return self.set_values(values)
 
-    def set_color_temp(self, color_temp, *, index=0, transition_time=None):
+    def set_color_temp(
+        self, color_temp: int, *, index: int = 0, transition_time: int | None = None
+    ) -> Command:
```
MartinHjelmare, line 159:
```suggestion
    ) -> Command[None]:
```

pytradfri/group.py
```diff
@@ -126,18 +167,24 @@ def set_color_temp(self, color_temp, *, index=0, transition_time=None):
 
         return self.set_values(values)
 
-    def set_hex_color(self, color, transition_time=None):
+    def set_hex_color(self, color: str, transition_time: int | None = None) -> Command:
```
MartinHjelmare, line 170:
```suggestion
    def set_hex_color(
        self, color: str, transition_time: int | None = None
    ) -> Command[None]:
```

pytradfri/group.py
```diff
@@ -126,18 +167,24 @@ def set_color_temp(self, color_temp, *, index=0, transition_time=None):
...
-        self, hue, saturation, brightness=None, *, index=0, transition_time=None
-    ):
+        self,
+        hue: int,
+        saturation: int,
+        brightness: int | None = None,
+        *,
+        index: int = 0,
+        transition_time: int | None = None,
+    ) -> Command:
```
MartinHjelmare, line 187:
```suggestion
    ) -> Command[None]:
```

pytradfri/group.py
```diff
@@ -153,7 +200,9 @@ def set_hsb(
 
         return self.set_values(values)
 
-    def set_xy_color(self, color_x, color_y, transition_time=None):
+    def set_xy_color(
+        self, color_x: int, color_y: int, transition_time: int | None = None
+    ) -> Command:
```
MartinHjelmare, line 205:
```suggestion
    ) -> Command[None]:
```

pytradfri/group.py
```diff
@@ -165,7 +214,9 @@ def set_xy_color(self, color_x, color_y, transition_time=None):
 
         return self.set_values(values)
 
-    def set_predefined_color(self, colorname, transition_time=None):
+    def set_predefined_color(
+        self, colorname: str, transition_time: int | None = None
+    ) -> Command:
```
MartinHjelmare, line 219:
```suggestion
    ) -> Command[None]:
```

pytradfri/resource.py
```diff
@@ -27,7 +30,7 @@ class ApiResource:
     """Base object for resources returned from the gateway."""
 
     _model_class: type[ApiResourceResponse] | None = None
-    raw: TypeRaw | ApiResourceResponse
+    raw: TypeRaw | ApiResourceResponse | GroupResponse
```
MartinHjelmare, line 33: Since GroupResponse inherits ApiResourceResponse, we don't need to change this.
```suggestion
    raw: TypeRaw | ApiResourceResponse
```

pytradfri/group.py
ggravlingen: I just created an empty group on my gateway and it looks like this:
```
    {
        "color_hex": null,
        "created_at": 1634990177,
        "dimmer": 0,
        "group_members": {
            "15002": {
                "9003": [
                    65555
                ]
            }
        },
        "id": 131083,
        "mood_id": "0",
        "name": "TRADFRI GROUP 9",
        "ota_update_state": null,
        "state": 0
    },
```

ggravlingen: Just posted here but I pasted the wrong group.

ggravlingen: I just did this: created a group in the app and queried the gateway. The group wasn't there. Hit back in the app and list all groups again. The new group is gone.

If I do the same and move a device into the grup it's not deleted and looks like this:
```
    {
        "color_hex": null,
        "created_at": 1646765851,
        "dimmer": 254,
        "group_members": {
            "15002": {
                "9003": [
                    65554
                ]
            }
        },
        "id": 131084,
        "mood_id": "0",
        "name": "Teatar",
        "ota_update_state": null,
        "state": 1
    }
```

The group I moved my device from looks like this:
```
    {
        "color_hex": null,
        "created_at": 1634661538,
        "dimmer": 0,
        "group_members": {
            "15002": {
                "9003": []
            }
        },
        "id": 131082,
        "mood_id": "0",
        "name": "Luftgrupp",
        "ota_update_state": null,
        "state": 0
    },
```

ggravlingen: So, I'm inclined to say that `group_members` will always have the keys `15002` and `9003`. The latter could be an empty list or a list of `int`.

MartinHjelmare: Sounds good! 👍

MartinHjelmare (approved):
Great!

PR #448, opened 2022-03-05
Make Command generic

MartinHjelmare:
By making the `Command` class a generic type we can increase the information of the api method signatures and command return types.

Note that this is technically a breaking change as I've removed the setter functionality of the `Command.result` attribute. It was making the typing confusing in the api modules, so I wanted to remove this and use a regular method call instead to process the result. In practice this should not be a breaking change since it's normally only the internal api methods that process command results.

I'll follow up with more changes in the api module to take advantage of the generic command type.

### DIFF
--- a/pytradfri/command.py
+++ b/pytradfri/command.py
@@ -1,13 +1,14 @@
 """Command implementation."""
 from __future__ import annotations
 
+from collections.abc import Callable
 from copy import deepcopy
-from typing import Any, Callable, Optional
+from typing import Any, Generic, TypeVar
 
-TypeProcessResultCb = Optional[Callable[[Any], Any]]
+T = TypeVar("T")
 
 
-class Command:
+class Command(Generic[T]):
     """The object for coap commands."""
 
     def __init__(
@@ -19,7 +20,7 @@ class Command:
         parse_json: bool = True,
         observe: bool = False,
         observe_duration: int = 0,
-        process_result: TypeProcessResultCb = None,
+        process_result: Callable[..., T] | None = None,
         err_callback: Callable[[Exception], None] | None = None,
     ) -> None:
         """Create object of class."""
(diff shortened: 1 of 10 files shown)


pytradfri/command.py
```diff
@@ -129,7 +125,7 @@ def combine_data(self, command2: Command | None) -> None:
             command2._data, self._data  # pylint: disable=protected-access
         )
 
-    def __add__(self, other: Command | None) -> Command:
+    def __add__(self, other: Command | None) -> Command[T]:
```
MartinHjelmare, line 128: Side note: I'm thinking maybe we should deprecate this feature and the combine command feature. We don't seem to use it in the library and I don't see why it should be needed by library users.

ggravlingen: I don't mind removing unused features to reduce maintenance load.

pytradfri/command.py
```diff
@@ -1,13 +1,14 @@
...
 from copy import deepcopy
-from typing import Any, Callable, Optional
+from typing import Any, Generic, TypeVar
 
-TypeProcessResultCb = Optional[Callable[[Any], Any]]
+T = TypeVar("T")
 
 
-class Command:
+class Command(Generic[T]):
```
ggravlingen, line 11: I have seen `T = TypeVar("T")` but never used it. Just did some reading of the concept and am I correct in understanding it as this? By using it, we can say that a method returns a `Command` of type `Group`. That automatically types the `process_result` as a callable taking argument `Group `and that the result is also of type `Group`?

ggravlingen: It sort of helps us do dynamic typing?

MartinHjelmare: `process_result` will return an instance of type T, yes.

It has two benefits as I see it.

1. It'll be easier to see what a Command class will return after the api has executed it. (this PR) 
2. We can use the TypeVar T of the command in the api and type the return value of the request method so that the request method will show the correct return value for any typed command we pass. (the follow up PR)

ggravlingen: I don't follow you completely but look forward to seeing the next PR to see what you mean.

ggravlingen (approved):

PR #451, opened 2022-03-07
Fix some typing issues

### LINKED ISSUE #449: Investigate types for path in Mood class
Following a discussion here, we should investigate why the properties are not of type `str`
https://github.com/home-assistant-libs/pytradfri/pull/446/files#r820148641

MartinHjelmare:
- We only use the `Command.path` property to construct the `Command.url` and there we copy all items to string. But I think it's more consistent if we make sure that the `path` items are strings from the beginning.

fixes #449

### DIFF
--- a/pytradfri/device/__init__.py
+++ b/pytradfri/device/__init__.py
@@ -71,14 +71,14 @@ class Device(ApiResource):
     raw: DeviceResponse
 
     @property
-    def application_type(self) -> int | None:
+    def application_type(self) -> int:
         """Return application type."""
         return self.raw.application_type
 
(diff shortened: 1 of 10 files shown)


ggravlingen:
Wouldn't this the issue I logged from the prematurely merged PR we discussed earlier? https://github.com/home-assistant-libs/pytradfri/issues/449?

ggravlingen (approved):

MartinHjelmare:
Yes, I forgot to link it. Done now.

ggravlingen:
Thanks!

PR #453, opened 2022-03-08
Cleanup mypy.ini

ggravlingen:
According to this line, we should be able to cleanup`mypy.ini` somewhat:
https://github.com/home-assistant-libs/pytradfri/blob/d16750d537830ea58e14bd45960106ea9ea5247f/mypy.ini#L38

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -70,29 +70,7 @@ no_implicit_optional = true
 warn_return_any = true
 warn_unreachable = true
 
-[mypy-pytradfri.api.__init__]
-check_untyped_defs = true
-disallow_incomplete_defs = true
-disallow_subclassing_any = true
-disallow_untyped_calls = true
-disallow_untyped_decorators = true
-disallow_untyped_defs = true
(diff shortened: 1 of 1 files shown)


mypy.ini
```diff
@@ -70,29 +70,7 @@ no_implicit_optional = true
...
-disallow_subclassing_any = true
-disallow_untyped_calls = true
-disallow_untyped_decorators = true
-disallow_untyped_defs = true
-no_implicit_optional = true
-warn_return_any = true
-warn_unreachable = true
-
-[mypy-pytradfri.device.controller]
+[mypy-pytradfri.api]
```
MartinHjelmare, line 73: I think we need a star pattern to glob all submodules in a directory. The libcoap api isn't typed so we're not globbing correctly here, since the build passes.

ggravlingen: You're right, and when adding * I also noticed we haven't typed `based_controller` properly. I'll close this PR and open again when we're through with that.

PR #454, opened 2022-03-08
Type hint base controller

ggravlingen:
It looks like there was a typo in `mypy.ini` causing us to not catch this file.

### DIFF
--- a/pytradfri/device/base_controller.py
+++ b/pytradfri/device/base_controller.py
@@ -1,7 +1,14 @@
 """Base class for a controller."""
 from __future__ import annotations
 
+from abc import abstractmethod
 from typing import TYPE_CHECKING
+
+from .air_purifier import AirPurifierResponse
+from .blind import BlindResponse
+from .light import LightResponse
+from .signal_repeater import SignalRepeaterResponse
+from .socket import SocketResponse
 
 if TYPE_CHECKING:
     # avoid cyclic import at runtime.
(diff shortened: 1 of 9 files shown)


pytradfri/device/base_controller.py
```diff
@@ -16,9 +23,13 @@ def __init__(self, device: Device) -> None:
         self._device = device
 
     @property
-    def raw(self):
+    @abstractmethod
+    def raw(
+        self,
+    ) -> list[BlindResponse] | list[SocketResponse] | list[
```
MartinHjelmare, line 29: Please extract the id attribute of the sub device response models to a parent model class that we define here in the base controller. Then inherit from this new parent class in all the sub device response models. Then set the new parent model response type as the list item return type here.

The base controller class shouldn't know any details of the sub device classes, only the common parts.

ggravlingen:
```
pytradfri/device/socket_control.py:14: error: Return type "List[SocketResponse]" of "raw" incompatible with return type "List[BaseResponse]" in supertype "BaseController"  [override]
```

Do we need to `cast`? It sounds by response that we shouldn't need to.

MartinHjelmare: Try using `Sequence` instead of `list` as container type.

https://docs.python.org/3/library/collections.abc.html#collections.abc.Sequence

ggravlingen: That seems to have worked! 💯

pytradfri/device/base_controller.py
```diff
@@ -1,13 +1,24 @@
...
+from pydantic import BaseModel, Field
+
+from ..const import ATTR_ID
 
 if TYPE_CHECKING:
     # avoid cyclic import at runtime.
     from . import Device
 
 
+class BaseResponse(BaseModel):
```
MartinHjelmare, line 16: Maybe move this to the resource module since we now use it in other modules than the device package?

ggravlingen: Makes sense

MartinHjelmare (approved):
Nice!

PR #455, opened 2022-03-09
Add type hints for util.py

ggravlingen:


pytradfri/util.py
```diff
@@ -19,7 +20,8 @@ def load_json(filename: str) -> list | dict:
     """
     try:
         with open(filename, encoding="utf-8") as fdesc:
-            return json.loads(fdesc.read())
+            return_value: list | dict = json.loads(fdesc.read())
```
ggravlingen, line 23: I tried, to no avail, using `cast(list|dict, json.loads(fdesc.read()))` here.

MartinHjelmare: When using `cast` we need to use `Union` instead of `|`.

Example:
https://github.com/home-assistant-libs/pytradfri/blob/bd6c39b01e3808cec576b399dc91e77851029ef5/pytradfri/api/aiocoap_api.py#L323

ggravlingen: Thanks!

MartinHjelmare (approved):

PR #456, opened 2022-03-10
Add type hints to smart_task.py

ggravlingen:


ggravlingen:
I'm breaking this down, probably class by class. I'm seeing code here that we really should deprecate and remove but let's do that after all typing is done.

pytradfri/smart_task.py
```diff
@@ -273,17 +275,24 @@ def __repr__(self):
 class StartActionItemController:
     """Class to edit settings for a task."""
 
-    def __init__(self, item, raw, state, path, devices_dict):
-        """Initialize TaskControl."""
+    def __init__(
+        self,
+        item: StartActionItem,
+        raw: TypeRaw,
+        state: int,
```
MartinHjelmare, line 282: There seem to be some ambiguity about the state type as there are two different origins of `StartActionItem`. For one state is a boolean and for the other it's an integer. A boolean is a subtype of an integer but we may need to convert it to a number if we expect that.

ggravlingen: Yes, this is a horrible mess (I don't blame anyone, I wrote this code when I just got into Python).

I believe we can drop this property as I can't see it being used anywhere?
https://github.com/home-assistant-libs/pytradfri/blob/421a9237a0ef231f18aceb79320595957d28fe8d/pytradfri/smart_task.py#L203

That will remove one of the places `StartActionItem` is being used.

ggravlingen: `state` is normally a `bool` throughout the code so I went for that option and converted them to an `int` where needed.

ggravlingen:
I touched a few lins that I hadn't planned on being in scope for this PR. It, hoever, felt motivated to make it clear what is going on.

MartinHjelmare:
Looks good. I think it may be easier to verify the correctness of the change if we type the whole module in one PR.

ggravlingen:
> Looks good. I think it may be easier to verify the correctness of the change if we type the whole module in one PR.

OK, I'll push more changes here.

tests/test_smart_task.py
```diff
@@ -14,6 +14,7 @@
         "15013": [
             {"5712": 18000, "5851": 254, "9003": 65537},
             {"5712": 18000, "5851": 254, "9003": 65538},
+            {"5712": 19000, "5851": 230, "9003": 65539},
```
ggravlingen, line 17: Wasn't sure if the `devices_list` property worked properly so I added one more record here.

pytradfri/smart_task.py
```diff
@@ -46,31 +54,71 @@
...
+    minute_start: int = Field(alias=ATTR_TIME_START_TIME_MINUTE)
+
+
+class RootStartActionResponse(SmartTaskMixin, BaseModel):
+    """Represent a smart action response."""
+
+    root_start_action: List[StartActionResponse] = Field(alias=ROOT_START_ACTION)
+
+
+class SmartTaskResponse(SmartTaskMixin, ApiResourceResponse):
```
ggravlingen, line 83: A sample response is available here:
https://github.com/home-assistant-libs/pytradfri/blob/master/tests/test_smart_task.py

pytradfri/smart_task.py
```diff
@@ -145,34 +191,38 @@ def __repr__(self):
...
+    def set_dimmer_start_time(self, hour: int, minute: int) -> Command[None]:
         """Set start time for task (hh:mm) in iso8601.
 
         NB: dimmer starts 30 mins before time in app
         """
         #  This is to calculate the difference between local time
         #  and the time in the gateway
-        d_now = self._gateway.get_gateway_info().current_time
+        gateway_info: Command[GatewayInfo] = self._gateway.get_gateway_info()
+        d_now = gateway_info.current_time  # type: ignore[attr-defined]
```
MartinHjelmare, line 219: This is a bug. We need to execute the command with an api before we can access the result.

pytradfri/smart_task.py
```diff
@@ -227,47 +284,53 @@ def __init__(self, task, index, state: bool, path, raw):
...
 
     @property
-    def transition_time(self):
+    def transition_time(self) -> int:
         """Transition runs for this long from the time in task_start.
 
         Value is in seconds x 10. Default to 0 if transition is missing.
         """
-        return self.raw.get(ATTR_TRANSITION_TIME, 0) / 60 / 10
+        return round(self.raw.transition_time / 60 / 10)
```
MartinHjelmare, line 321: Are we sure that transition_time is not optional?

See https://github.com/home-assistant-libs/pytradfri/pull/276

ggravlingen: Oh, I looked at the sample data in the test folder. But you have digged into this more recently so I believe you are right.

pytradfri/smart_task.py
ggravlingen: Hm, do we really want an API call here? One of the features of this library is to not have API calls in the code representing the gateway's responses?

ggravlingen: (I understand that it's a bug)

ggravlingen: Perhaps a way forward is to assume the gateway has the same local time as the environment where the script is run?

MartinHjelmare: Maybe add a method, eg `calibrate_time`, that returns a Command with a result callback that sets the time diff between local time and gateway time as an attribute on the smart task?

We default this attribute to timedelta(0).

ggravlingen: Tried implementing something along the line of your comments here

pytradfri/smart_task.py
```diff
@@ -177,6 +179,15 @@ def task_control(self) -> TaskControl:
         """Control a task."""
         return TaskControl(self, self.state, self.path, self._gateway)
 
+    @property
+    def delta_time_gateway_local(self) -> timedelta:
+        """Return difference between local time and time set on gateway."""
+        return timedelta(0)
+
+    @delta_time_gateway_local.setter
```
MartinHjelmare, line 188: Just make it an attribute. I don't think we need a property.

pytradfri/smart_task.py
```diff
@@ -208,25 +219,36 @@ def tasks(self) -> list[StartActionItem]:
             for idx in range(len(self.raw.root_start_action))
         ]
 
+    def calibrate_time(self) -> Command[None]:
+        """Calibrate difference between local time and gateway time."""
+
+        def process_result(result: TypeRaw) -> None:
+            gateway_info: GatewayInfo = GatewayInfo(result)
+            if gateway_info.current_time is not None:
```
MartinHjelmare, line 228: We can invert the condition and return if true. Then outdent below.

pytradfri/smart_task.py
ggravlingen: I pushed before I saw this comment, will refactor.

pytradfri/smart_task.py
```diff
@@ -225,12 +216,14 @@ def calibrate_time(self) -> Command[None]:
...
         def process_result(result: TypeRaw) -> None:
             gateway_info: GatewayInfo = GatewayInfo(result)
-            if gateway_info.current_time is not None:
+            if gateway_info.current_time:
                 d_now: dt = gateway_info.current_time
                 d_utcnow: dt = dt.utcnow()
                 diff: timedelta = d_now - d_utcnow
 
-                self._task.delta_time_gateway_local = diff
+                setattr(self._task, "delta_time_gateway_local", diff)
```
MartinHjelmare, line 224:
```suggestion
                self._task.delta_time_gateway_local = diff
```

pytradfri/smart_task.py
```diff
@@ -103,7 +103,7 @@ def __init__(self, gateway: Gateway, raw: TypeRaw) -> None:
         """Initialize the class."""
         super().__init__(raw)
         self._gateway = gateway
-        self._delta_time_gateway_local: timedelta = timedelta(0)
+        self.delta_time_gateway_local: timedelta = timedelta(0)
```
MartinHjelmare, line 106: Do we need the type annotation?

ggravlingen: We did not, at least mypy didn't react to me removing it.

pytradfri/smart_task.py
```diff
@@ -143,80 +194,104 @@ def __repr__(self):
...
+        def process_result(result: TypeRaw) -> None:
+            gateway_info: GatewayInfo = GatewayInfo(result)
+            if gateway_info.current_time:
+                d_now: dt = gateway_info.current_time
+                d_utcnow: dt = dt.utcnow()
+                diff: timedelta = d_now - d_utcnow
+
+                self._task.delta_time_gateway_local = diff
+
+            return None
```
MartinHjelmare, line 226: We don't need an explicit return statement.

pytradfri/smart_task.py
```diff
@@ -143,80 +194,104 @@ def __repr__(self):
...
+            for idx in range(len(self.raw.root_start_action))
         ]
 
-    def set_dimmer_start_time(self, hour, minute):
+    def calibrate_time(self) -> Command[None]:
+        """Calibrate difference between local time and gateway time."""
+
+        def process_result(result: TypeRaw) -> None:
+            gateway_info: GatewayInfo = GatewayInfo(result)
+            if gateway_info.current_time:
```
MartinHjelmare, line 219: I'd invert this check and `return` if it's true. Then we can outdent below.

ggravlingen: Sorry, realise now that didn't quite understand your comment for this yesterday. Is this what you meant?

pytradfri/smart_task.py
```diff
@@ -216,14 +216,14 @@ def calibrate_time(self) -> Command[None]:
...
             gateway_info: GatewayInfo = GatewayInfo(result)
-            if gateway_info.current_time:
-                d_now: dt = gateway_info.current_time
-                d_utcnow: dt = dt.utcnow()
-                diff: timedelta = d_now - d_utcnow
+            if not gateway_info.current_time:
+                return
 
-                self._task.delta_time_gateway_local = diff
+            d_now: dt = gateway_info.current_time
```
MartinHjelmare, line 222: Do we need these type annotations? Normally mypy will be able to read the types from the objects. It's only when the object is unknown or generic we may need to help mypy and define the type explicitly.

pytradfri/smart_task.py
MartinHjelmare: Yes. 👍

pytradfri/smart_task.py
ggravlingen: Looks like no, I'll remove them.

MartinHjelmare (approved):
Great!

PR #461, opened 2022-03-15
Enforce strict typing

ggravlingen:


ggravlingen:
https://github.com/home-assistant-libs/pytradfri/blob/3f8573fa6a9b2dca16618b158fa137f68c5bcbdb/pytradfri/api/aiocoap_api.py#L56

What should be the type hint here? In a previous PR, you wrote `The Task should be typed with its result value.` but I don't quite understand what you mean here.

ggravlingen:
When this PR is merged, I believe we only have the libcoap-file to go before it's fully typed?

pytradfri/command.py
```diff
@@ -119,15 +119,15 @@ def _merge(
                 b_dict[key] = value
         return b_dict
 
-    def combine_data(self, command2: Command | None) -> None:
+    def combine_data(self, command2: Command[T] | None) -> None:
```
MartinHjelmare, line 122: I'm not sure that the parameter is limited to a command with the same return value as this command. Maybe put `Any` instead of `T` for now.

ggravlingen:
Should we also add type hints to `__main__.py`?

pytradfri/command.py
```diff
@@ -119,15 +119,15 @@ def _merge(
...
+    def combine_data(self, command2: Command[Any] | None) -> None:
         """Combine data for this command with another."""
         if command2 is None:
             return
         self._data = self._merge(
             command2._data, self._data  # pylint: disable=protected-access
         )
 
-    def __add__(self, other: Command | None) -> Command[T]:
+    def __add__(self, other: Command[T] | None) -> Command[T]:
```
MartinHjelmare, line 130:
```suggestion
    def __add__(self, other: Command[Any] | None) -> Command[T]:
```

MartinHjelmare:
> Should we also add type hints to `__main__.py`?

Yes, we should.

pytradfri/util.py
```diff
@@ -13,14 +13,14 @@
...
-def load_json(filename: str) -> list | dict:
+def load_json(filename: str) -> list[Any] | dict[Any, Any]:
     """Load JSON data from a file and return as dict or list.
 
     Defaults to returning empty dict if file is not found.
     """
     try:
         with open(filename, encoding="utf-8") as fdesc:
-            return cast(Union[dict, list], json.loads(fdesc.read()))
+            return cast(Union[dict[Any, Any], list[Any]], json.loads(fdesc.read()))
```
MartinHjelmare, line 23:
```suggestion
            return cast(Union[Dict[Any, Any], List[Any]], json.loads(fdesc.read()))
```

ggravlingen:
> > Should we also add type hints to `__main__.py`?
> 
> Yes, we should.

Great, I’ll look into that.

pytradfri/util.py
```diff
@@ -3,7 +3,7 @@
 
 import json
 import logging
-from typing import Iterator, Union, cast
+from typing import Any, Iterator, Union, cast
```
MartinHjelmare, line 6:
```suggestion
from typing import Any, Dict, Iterator, List, Union, cast
```

pytradfri/util.py
ggravlingen: What is the reason for lowercase `dict` not working here?

MartinHjelmare: The future annotations doesn't work within `cast`.

ggravlingen: Thanks!

pytradfri/api/aiocoap_api.py
```diff
@@ -320,4 +322,4 @@ def _process_output(res: Message, parse_json: bool = True) -> list | dict | str
     if not parse_json:
         return output
 
-    return cast(Union[list, dict], json.loads(output))
+    return cast(Union[list[Any], dict[Any, Any]], json.loads(output))
```
MartinHjelmare, line 325: We need the same adjustment here.

MartinHjelmare (approved):

MartinHjelmare:
The api task returns a `Context` instance.

ggravlingen:
I made a change (added `Context` as return type) after your approval so I won't merge. Please do if the change was acceptable.

MartinHjelmare (approved):
Great!

PR #462, opened 2022-03-16
Add type hints for libcoap.py

ggravlingen:


pytradfri/api/libcoap_api.py
```diff
@@ -206,22 +221,5 @@ def _process_output(output, parse_json=True):
     if output.startswith(SERVER_ERROR_PREFIX):
         raise ServerError(output)
     if not parse_json:
-        return output
-    return json.loads(output)
-
-
-def retry_timeout(api, retries=3):
```
ggravlingen, line 213: I can't see that this is being used anywhere? I thus removed it. Maybe do that in a separate PR and then look at adding type hints?

pytradfri/api/libcoap_api.py
```diff
@@ -21,25 +22,34 @@
...
 
     @psk.setter
-    def psk(self, value):
+    def psk(self, value: str) -> None:
         """Set psk."""
         self._psk = value
 
     def _base_command(self, method: str) -> list[str]:
         """Return base command."""
+        if self._psk is None:
```
ggravlingen, line 50: Not sure about this one, but `_psk` will not always be a string so we have to handle that somehow.

pytradfri/api/libcoap_api.py
```diff
@@ -98,7 +110,9 @@ def _execute(self, api_command, *, timeout=None):
         api_command.process_result(_process_output(return_value, parse_json))
         return api_command.result
 
-    def request(self, api_commands, *, timeout=None):
+    def request(
+        self, api_commands: list[Command[T]], *, timeout: int | None = None
+    ) -> T | list[T] | list[Any] | None:
         """Make a request. Timeout is in seconds."""
         if not isinstance(api_commands, list):
```
ggravlingen, line 117: Do we need this? Don't we expect api_commands to always be a list given the type hint for `api_commands`?

pytradfri/api/libcoap_api.py
MartinHjelmare: Raise a RuntimeError. The user needs to generate a key first.

pytradfri/api/libcoap_api.py
MartinHjelmare: Check the typing of the async api. It should be the same, including overrides.

pytradfri/api/libcoap_api.py
MartinHjelmare: I think we can remove it. I think it was used a long time ago as a helper.

pytradfri/api/libcoap_api.py
ggravlingen: I add to make some adjustments as `_execute` can also return `None`.

MartinHjelmare: No, execute should always return a command result. When starting an observation we should also return the command result.

pytradfri/api/libcoap_api.py
```diff
@@ -180,15 +206,18 @@ def generate_psk(self, security_key):
             self._psk = security_key
 
             # Ask the Gateway to generate the psk for the identity.
-            self._psk = self.request(Gateway().generate_psk(existing_psk_id))
+            command: list[Command[str]] = [Gateway().generate_psk(existing_psk_id)]
```
MartinHjelmare, line 209: Revert this change.

MartinHjelmare (approved):
Great!

Issue #457, opened 2022-03-13
Suggestions for deprecations


ggravlingen:
Listing the things we've been discussing to deprecate, and adding a few suggestions on my own:
- `Command._add__`: not in use?
- `Command.combine_data` not in use?
- `smart_task.StartAction.devices`: can't see this being used anywhere, nor that it adds any functionality?
- `libcoap_api`: do we really need to keep maintaining this?

ggravlingen:
Ping @MartinHjelmare for discussion.

MartinHjelmare:
What does the `ROOT_START_ACTION` item mean?

I think it could be ok to deprecate the libcoap api. Although currently it doesn't add very much maintenance burden either think. So I would recommend to keep it.

PR #463, opened 2022-03-18
Add type hints to __main__.py

ggravlingen:
This PR adds type hints to `__main__.py`.


pytradfri/__main__.py
```diff
@@ -14,9 +18,21 @@
 CONFIG_FILE = "tradfri_standalone_psk.conf"
 
 
-def main(org_args):
-    """Run main."""
-    global devices, homekit_id, light, api, lights, gateway, groups, moods, tasks, dump_devices, dump_all  # pylint: disable=global-variable-undefined, invalid-name
+if __name__ == "__main__":
```
ggravlingen, line 21: By putting all the code here instead of in a function, we don't need the `global` definition, which was messing up the type hints.

ggravlingen:
Deleting the tox.ini was not intended. Will put it back.

ggravlingen:
I believe this concludes the task of adding type hints to the library? I remember you did a breaking change in one of the pull requests, albeit a small one. Should I bump to 10.0.0 or is this 9.1.0? If we're going for 10 we might as well drop the methods discussed in https://github.com/home-assistant-libs/pytradfri/issues/457.

MartinHjelmare (approved):
In a follow up, we can make `_model_class` non optional and remove the cases for it in resource.py.

We can also now mark the library as typed with a file `py.typed`.

MartinHjelmare:
I think we can make a major bump after removing the stale methods.

PR #464, opened 2022-03-19
Make ApiResourceResponse mandatory and mark library as typed

ggravlingen:


MartinHjelmare (commented):
The py.typed file should go inside the pytradfri package. Also make sure it's included when building the package.

Example:
https://github.com/MartinHjelmare/aiovlc/blob/a1cfe347b397632d03637c76ffe2e8f786b2f68d/setup.py#L24-L25

pytradfri/resource.py
```diff
@@ -31,41 +31,27 @@ class ApiResourceResponse(BaseResponse):
...
 
     @property
     def id(self) -> int:
         """Id."""
-        if self._model_class:
-            resource_id = self.raw.id  # type: ignore[union-attr]
-        else:
-            resource_id = self.raw[ATTR_ID]  # type: ignore[index]
-        return resource_id
+        return self.raw.id  # type: ignore[union-attr]
```
MartinHjelmare, line 44: We shouldn't need the ignores anymore now.

MartinHjelmare:
Please move `py.typed` under `pytradfri` dir.

MartinHjelmare (approved):

PR #465, opened 2022-03-19
Remove unused code

ggravlingen:
Closes #463


tests/test_smart_task.py
```diff
@@ -98,28 +98,6 @@ def test_smart_task_info():
     assert task.dimmer == 254
 
 
-def test_smart_task_set_start_action_dimmer():
-    """Test start dimmer."""
-    gateway = Gateway()
-
-    cmd = (
-        SmartTask(gateway, TASK).start_action.devices[0].item_controller.set_dimmer(30)
```
ggravlingen, line 106: This is the only place where I have seen `devices` been used. I feel it's based on an incorrect assumption that you would want to control a device from a smart task.

pytradfri/command.py
```diff
@@ -119,25 +118,6 @@ def _merge(
                 b_dict[key] = value
         return b_dict
```
MartinHjelmare, line 119: We can remove `_merge` too I think, right?

ggravlingen: Yes, it was being used in combine only.

tests/test_smart_task.py
MartinHjelmare: This is calling the `set_dimmer` method in the smart_task module. But it looks like you can do the same via `task_control.tasks`. Is there a difference in what `StartActionItem` instances are enumerated in the two different lists?

ggravlingen: To be honest, I can't remember what caused me to go for this ambiguous design pattern.  The order seem to have been the same in both occurances:

TaskControl:
```python
return [
  StartActionItem(self._task, idx, self.state, self.path, self.raw)
  for idx in range(len(self.raw.root_start_action))
]`

StartAction:
```python
[
  StartActionItem(self.smart_task, i, self.state, self.path, self.raw)
  for i in range(len(self.raw.root_start_action))
]
```

MartinHjelmare: Then I think we can remove this as there's an alternative in place.

MartinHjelmare (approved):

PR #467, opened 2022-03-19
Add checks to pre-commit

ggravlingen:
I accidentally committed and pushed the version bump to master (have reverted that). Don't want to do the same mistake again so I'm adding a rule to pre-commit.

On a side note, I use Azure's git repos for work and we can set a rule preventing commits without PRs. Can that be on Github too? I'm not an admin of this repo so I can't check.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
       - id: isort
         name: isort
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.2.0
+    rev: v4.1.0
     hooks:
       - id: requirements-txt-fixer
       - id: check-json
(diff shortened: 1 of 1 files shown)


MartinHjelmare:
We can set rules for branches, yes.

.pre-commit-config.yaml
```diff
@@ -24,6 +24,15 @@ repos:
     hooks:
       - id: isort
         name: isort
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.1.0
+    hooks:
+      - id: requirements-txt-fixer
+      - id: check-json
```
MartinHjelmare, line 31: What do these two hooks do?

ggravlingen: requirements-txt-fixer - sorts entries in requirements.txt.

check-json - checks json files for parseable syntax.

MartinHjelmare (approved):
Looks good!

PR #480, opened 2022-04-22
Clarify meaning of filter_status property

ggravlingen:
Six months have passed since I installed the filter on the device and the `filter_lifetime_remaining` property now reports a negative value of 31 minutes. In the app, there is an indicator that the filter expires today.

I therefore propose this clarification to the `filter_status` property.

```
DeviceResponse(
  id=65554,
  name='Luftrenare',
  created_at=1634661517,
  ota_update_state=0,
  air_purifier_control=[
    AirPurifierResponse(
      id=0,
      air_quality=11,
      controls_locked=0,
      fan_speed=10,
      filter_lifetime_remaining=-31,
      filter_lifetime_total=259200,
      filter_runtime=259231,
      filter_status=1,
      leds_off=0,
      mode=1,
      motor_runtime_total=259231
   )
],

application_type=10,
blind_control=None,
device_info=DeviceInfoResponse(manufacturer='IKEA of Sweden', model_number='STARKVIND Air purifier', serial='',firmware_version='1.0.033', power_source=1, battery_level=None),
last_seen=1650565905,
light_control=None,
reachable=1,
signal_repeater_control=None,
socket_control=None
)
```

### DIFF
--- a/pytradfri/device/air_purifier.py
+++ b/pytradfri/device/air_purifier.py
@@ -93,7 +93,7 @@ class AirPurifier:
 
     @property
     def filter_status(self) -> bool:
-        """Return true if filter needs to be replaced."""
+        """Return True if filter needs to be replaced."""
         return bool(self.raw.filter_status)
 
     @property
--- a/pytradfri/device/air_purifier.py
+++ b/pytradfri/device/air_purifier.py
@@ -92,9 +92,9 @@ class AirPurifier:
         return self.raw.filter_runtime
 
     @property
-    def filter_status(self) -> int:
-        """Return filter status."""
-        return self.raw.filter_status
+    def filter_status(self) -> bool:
+        """Return true if filter needs to be replaced."""
+        return bool(self.raw.filter_status)
 
     @property
     def is_auto_mode(self) -> bool:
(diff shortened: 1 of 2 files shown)


ggravlingen:
Just noticed there was a red light on the purifier which normally isn't there. After replacing the filter and pressing the reset-button, the red light is no longer lit.

pytradfri/device/air_purifier.py
```diff
@@ -92,9 +92,13 @@ def filter_runtime(self) -> int:
...
-    def filter_status(self) -> int:
-        """Return filter status."""
-        return self.raw.filter_status
+    def filter_status(self) -> bool:
+        """
+        Return True if filter needs to be replaced.
+
+        This property is true when filter_lifetime_remaining is less than zero.
+        """
+        return bool(self.raw.filter_status)
```
MartinHjelmare, line 101: This is a breaking change although minor I would say.

MartinHjelmare (approved):

pytradfri/device/air_purifier.py
ggravlingen: So, a major version bump in a follow up?

MartinHjelmare: Yeah, good to stay consistent with semver.

PR #495, opened 2022-05-28
Track updates of pre-commit hooks

ggravlingen:
I had a look to see if dependabot could keep the pre-commit hooks update. Apparently it wasn't in their roadmap, but I found this instead:
https://browniebroke.com/blog/gh-action-pre-commit-autoupdate/

### DIFF
--- a/.github/workflows/pre-commit-update.yml
+++ b/.github/workflows/pre-commit-update.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v2
         with:
-          token: ${{ secrets.CPR_GITHUB_TOKEN }}
+          token: ${{ secrets.GITHUB_TOKEN }}
           branch: update/pre-commit-autoupdate
           title: Auto-update pre-commit hooks
           commit-message: Auto-update pre-commit hooks
--- a/.github/workflows/pre-commit-update.yml
+++ b/.github/workflows/pre-commit-update.yml
@@ -1,0 +1,30 @@
+name: Pre-commit auto-update
+on:
+  schedule:
+    # Run on mondays
+    - cron: "0 0 * * 1"
+
+jobs:
+  auto-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: Install pre-commit
+        run: pip install pre-commit
+      - name: Run pre-commit autoupdate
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-commit-update.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          python-version: 3.8
+      - name: Install pre-commit
+        run: pip install pre-commit
+      - name: Run pre-commit autoupdate
+        run: pre-commit autoupdate
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v2
+        with:
+          token: ${{ secrets.CPR_GITHUB_TOKEN }}
```
MartinHjelmare, line 23:
```suggestion
          token: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/pre-commit-update.yml
```diff
@@ -0,0 +1,30 @@
...
+on:
+  schedule:
+    # Run on mondays
+    - cron: "0 0 * * 1"
+
+jobs:
+  auto-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
MartinHjelmare, line 11: Please pin all action versions to the full semantic version.
```suggestion
      - uses: actions/checkout@v2.0.0
```

.github/workflows/pre-commit-update.yml
```diff
@@ -0,0 +1,30 @@
...
+    # Run on mondays
+    - cron: "0 0 * * 1"
+
+jobs:
+  auto-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.0.0
+      - name: Set up Python
+        uses: actions/setup-python@v2
```
MartinHjelmare, line 13: Please update all version tags to the full semantic version.
```suggestion
        uses: actions/setup-python@v2.0.0
```

.github/workflows/pre-commit-update.yml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  auto-update:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.0.0
+      - name: Set up Python
+        uses: actions/setup-python@v2.0.0
+        with:
+          python-version: 3.8
```
ggravlingen, line 15: Pin minor here too?

MartinHjelmare: No, we don't do that.

MartinHjelmare (approved):
Looks good!

PR #510, opened 2022-06-23
Add APIRequestProtocol for APIfactory.request

MartinHjelmare:
- A protocol is needed to allow users of the library to type the `APIFactory.request` method which is the main api function in use.


ggravlingen:
So essentially, the user would import the protocol and use it as a type hint in their application?

MartinHjelmare:
Yeah. I'm working on bumping the library in Home Assistant and this is one of the things that we need to address.

ggravlingen:
Great initiative!

ggravlingen (approved):

PR #511, opened 2022-06-23
Rework light supported features

MartinHjelmare:
- The "can_set_*" attributes on the light control instance were broken after the big refactor with Pydantic.
- Fix that by introducing a new set of properties on the Light class that returns the available features of the light.
- I think these properties belong on the Light class instead of LightControl as the LightResponse data is needed to check supported features and that data is available directly on the Light instance but only indirectly on the LigthControl instance. It doesn't matter for the LightControl instance what kind of lights it controls.

### DIFF
--- a/pytradfri/color.py
+++ b/pytradfri/color.py
@@ -46,16 +46,16 @@ def supported_features(data: LightResponse) -> int:
     """Return supported features."""
     supported_color_features = 0
 
-    if data.dimmer:
+    if data.dimmer is not None:
         supported_color_features = supported_color_features + SUPPORT_BRIGHTNESS
 
     if data.color_hex:
         supported_color_features = supported_color_features + SUPPORT_HEX_COLOR
 
-    if data.color_mireds:
+    if data.color_mireds is not None:
         supported_color_features = supported_color_features + SUPPORT_COLOR_TEMP
 
-    if data.color_xy_x and data.color_xy_y:
+    if None not in (data.color_xy_x, data.color_xy_y):
         supported_color_features = supported_color_features + SUPPORT_XY_COLOR
 
     if (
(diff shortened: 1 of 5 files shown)


pytradfri/color.py
```diff
@@ -46,16 +46,16 @@ def supported_features(data: LightResponse) -> int:
     """Return supported features."""
     supported_color_features = 0
 
-    if data.dimmer:
+    if data.dimmer is not None:
```
MartinHjelmare, line 49: The integer 0 is evaluated as falsy, so we need to check for None explicitly for the attributes that can return integers.

pytradfri/color.py
```diff
@@ -46,16 +46,16 @@ def supported_features(data: LightResponse) -> int:
...
         supported_color_features = supported_color_features + SUPPORT_BRIGHTNESS
 
     if data.color_hex:
```
ggravlingen, line 52: Color_hex is a string and thus truthy here, ie we don't need to check for None?

MartinHjelmare: I'm assuming that an empty string would be an error/incorrect. But we can change to check for None if we want to allow an empty string.

ggravlingen: Hmm, I'm not sure here. If I remember correctly, the gateway has a set of permitted hex colors. If one of these colors is set, `color_hex` will contain a hex value. If a color that does not have a string representation is set, then `color_hex` takes some other value (might be "0" as seen here https://github.com/home-assistant-libs/pytradfri/blob/b61737b5241ece653f89f554df7b386d1eb4cb49/tests/devices.py#L31).

I'm inclined to believe that any color bulb would support hex but again, I'm not sure.

MartinHjelmare: Ok. I think we can keep it like this for now. If we know better in the future we can update it.

ggravlingen (approved):

ggravlingen:
I agree with the changes, makes the code a lot cleaner. Thanks!

PR #513, opened 2022-06-25
Remove support for Python 3.8

MartinHjelmare:
- I suggest we drop support for Python 3.8.
- Home Assistant only supports Python 3.9 and Python 3.10.
- aiocoap is thinking about dropping Python 3.8.
- It simplifies typing somewhat (will address in separate PR that solves pylint issues).


ggravlingen (approved):
Sounds good!

PR #514, opened 2022-06-26
Address remaining pylint issues in pytradfri

MartinHjelmare:
- This solves the pylint issues in the pytradfri package.
- I'll follow up with more PRs for solving pylint for examples and tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,6 +68,7 @@ disable = [
     "format",
 #    "abstract-class-little-used",
     "abstract-method",
+    "consider-alternative-union-syntax",
     "consider-using-assignment-expr",
     "cyclic-import",
     "duplicate-code",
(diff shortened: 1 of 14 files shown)


pyproject.toml
```diff
@@ -68,6 +68,7 @@ disable = [
     "format",
 #    "abstract-class-little-used",
     "abstract-method",
+    "consider-alternative-union-syntax",
```
MartinHjelmare, line 71: When Python 3.10 is the minimum version (or pydantic is updated to support the new union syntax with Python 3.9) we can remove this disable and change to the new syntax.

ggravlingen: Maybe add this as a comment in the code for future reference?

ggravlingen (approved):
Other than my one small questions it looks good to me.

PR #543, opened 2022-08-28
Move mypy config to pyproject.toml

ggravlingen:
This PR suggests moving the mypy config to pyproject.toml. The purpose is to reduce the clutter of having config spread out over too many files.


pyproject.toml
```diff
@@ -95,3 +95,29 @@ norecursedirs = [
...
+disallow_untyped_decorators = true
+disallow_untyped_defs = true
+no_implicit_optional = true
+warn_return_any = true
+warn_unreachable = true
+plugins = "pydantic.mypy"
+strict = true
+
+[[tool.mypy.overrides]]
+module = 'tests.*'
```
MartinHjelmare, line 122:
```suggestion
module = "tests.*"
```

MartinHjelmare (approved):
Thanks!

PR #544, opened 2022-08-28
Make flake8 ignore settings less lax

ggravlingen:
We can quite easiliy remove these ignores so I propose we do that.


MartinHjelmare (approved):
Thanks!

PR #614, opened 2023-01-16
Fix smart task start action response dimmer attribute

blast-hardcheese:
Resolves #610

Also fleshed out some examples for how to interact with `SmartTask`s, happy to revert some or all of those if undesired.

Thanks for this library!

PR #615, opened 2023-01-16
Add set_repeat_days to SmartTask

blast-hardcheese:
- Added `Days` with constants into `const.py`
- Added `TaskControl.set_repeat_days` for SmartTask
- Added `lights` to the onboarding helper text
- Added `sockets` and `socket` to `__main__`
- Added an example text for how to use `set_repeat_days` in `__main__`

### DIFF
--- a/pytradfri/__main__.py
+++ b/pytradfri/__main__.py
@@ -138,6 +138,9 @@ if __name__ == "__main__":
     if socket:
         print("> sockets")
         print("> socket.socket_control.sockets")
+        print("> from pytradfri.const.Days import TUESDAY, THURSDAY, SATURDAY, SUNDAY")
+        print("> days = TUESDAY | THURSDAY | SATURDAY | SUNDAY")
+        print("> api(socket.socket_control.set_repeat_days(days))")
     if tasks:
         print("> tasks[0].repeat_days_list")
     print("> api(gateway.reboot())")
--- a/pytradfri/__main__.py
+++ b/pytradfri/__main__.py
@@ -129,6 +129,7 @@ if __name__ == "__main__":
     print("> devices")
     print("> homekit_id")
     if light:
+        print("> lights")
         print("> light.light_control.lights")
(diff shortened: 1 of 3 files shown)


blast-hardcheese:
In response to ["We don't use the const module above. I think we should be consistent"](https://github.com/home-assistant-libs/pytradfri/pull/614#discussion_r1071466200),

I think having symbolic names for the individual days is critical as a disambiguation, since "`0b1000000` is Monday" isn't obvious, and avoiding trial and error seems wise

blast-hardcheese:
Actually, I didn't see `WEEKDAYS` initially. If there's some other way you'd like me to approach this, I'm happy to do so

pytradfri/__main__.py
```diff
@@ -122,12 +129,19 @@ def dump_devices() -> None:
...
         print("> api(light.light_control.set_dimmer(254))")
         print("> api(light.light_control.set_xy_color(30015, 26870))")
         print("> api(light.light_control.set_predefined_color('Warm Amber'))")
         print("> api(lights[1].light_control.set_dimmer(20))")
+    if socket:
+        print("> sockets")
+        print("> socket.socket_control.sockets")
+        print("> from pytradfri.const.Days import TUESDAY, THURSDAY, SATURDAY, SUNDAY")
+        print("> days = TUESDAY | THURSDAY | SATURDAY | SUNDAY")
+        print("> api(socket.socket_control.set_repeat_days(days))")
```
blast-hardcheese, line 144: This would avoid the modification to `const.py`:

```suggestion
        print("> from pytradfri.smart_task import WEEKDAYS")
        print("> days = WEEKDAYS.tue | WEEKDAYS.thur | WEEKDAYS.sat | WEEKDAYS.sun")
        print("> api(socket.socket_control.set_repeat_days(days))")
```

MartinHjelmare: Please use the existing constant in `smart_task`.

PR #626, opened 2023-02-05
Add Python 3.11 to CI/CD pipeline

ggravlingen:
I read that support for Python 3.9 has been dropped in core so I guess we could drop it here too. Before doing that, I propose we add 3.11 to the pipeline run.


MartinHjelmare (approved):
Thanks!

PR #633, opened 2023-02-12
Use Python 3.10 syntax

ggravlingen:
I ran `pyupgrade` on the code base to update it to use Python 3.10 syntax.


MartinHjelmare (approved):
Thanks!

PR #635, opened 2023-02-12
Fix overgeneral-exceptions error

ggravlingen:
Pylint flooded output with `pylint: Command line or configuration file:1: UserWarning: Specifying exception names in the overgeneral-exceptions option without module name is deprecated and support for it will be removed in pylint 3.0. Use fully qualified name (maybe 'builtins.Exception' ?) instead.` before making this adjustment.


MartinHjelmare (approved):
Thanks!

PR #638, opened 2023-02-13
Add quotes to Python version in mypy config

ggravlingen:
Resolves this error `pyproject.toml: [mypy]: python_version: Python 3.1 is not supported (must be 3.4 or higher). You may need to put quotes around your Python version`


MartinHjelmare (approved):
Thanks!

PR #646, opened 2023-03-03
Add basic support for ruff

ggravlingen:
I noticed `core` has started using this and I've done some testing in other repos during the week. It's really neat so I suggest we use it here too.

Adding basic support here and enabling a few rules.

### DIFF
--- a/pytradfri/__main__.py
+++ b/pytradfri/__main__.py
@@ -3,8 +3,8 @@ from __future__ import annotations
 
 import argparse
 import logging
-from pprint import pprint
 import uuid
+from pprint import pprint
 
 from pytradfri.api.libcoap_api import APIFactory
 from pytradfri.device import Device
--- a/pytradfri/__main__.py
+++ b/pytradfri/__main__.py
@@ -3,8 +3,8 @@ from __future__ import annotations
 
 import argparse
 import logging
-import uuid
 from pprint import pprint
+import uuid
 
 from pytradfri.api.libcoap_api import APIFactory
 from pytradfri.device import Device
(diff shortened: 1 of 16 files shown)


ggravlingen:
Having thought about it a bit, I guess we can remove `flake8` from the pre-commit rules and the CI/CD?

MartinHjelmare:
Yes, I think we can remove flake8 and maybe isort too if we activate that in ruff?

pytradfri/__main__.py
```diff
@@ -3,8 +3,8 @@
 
 import argparse
 import logging
-from pprint import pprint
 import uuid
+from pprint import pprint
```
MartinHjelmare, line 7: Can we add isort configuration to pyproject.toml to avoid changing the sorting?

MartinHjelmare: I think we may just need to make sure that ruff picks up the existing isort config?

ggravlingen: > Can we add isort configuration to pyproject.toml to avoid changing the sorting?

Yes, I'll add that.

ggravlingen: I thought ruff would extend config automatically. Maybe there's a flag for that?

ggravlingen: I did some digging in ruff's docs and GitHub issues section and couldn't find any obvious solutions but adding a separate section.

MartinHjelmare: Ok.

MartinHjelmare (approved):

PR #729, opened 2023-07-22
Refactor test and QA pipeline

ggravlingen:
This PR:
- Separates linting and unit tests into separate jobs
- Drops support for `tox`
- Posts coverage to the pull request


.github/workflows/test.yml
```diff
@@ -1,31 +1,148 @@
...
+      python_versions: ${{ steps.info.outputs.python_versions }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3.5.3
+
+      - name: Filter for core changes
+        uses: dorny/paths-filter@v2.11.1
+        id: changes
+        with:
+          filters: |
```
MartinHjelmare, line 36: We don't seem to need the filter as we always run either all jobs or no jobs, besides the info job. Also, this project isn't large enough to justify a filter I think.

Caching the environment setup is good as installing the requirements without cache takes a minute.

ggravlingen: A valid point, I'll remove the filter.

.github/workflows/test.yml
```diff
@@ -1,31 +1,130 @@
-# This workflow will install Python dependencies, run tests and lint with a single version of Python
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-python-with-github-actions
-
 name: Run Tests
 
+# Explicitly set permissions to allow Dependabot workflow runs to write in the PR
+# for coverage's reporting.
+# By default, these are read-only when the actions are ran by Dependabot
+# https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#changing-github_token-permissions
+permissions: write-all
```
MartinHjelmare, line 7: I think we should use codecov instead. See eg the zwave-js-server-python project.

.github/workflows/test.yml
```diff
@@ -1,31 +1,130 @@
...
+
+env:
+  ALL_PYTHON_VERSIONS: "['3.10', '3.11']"
```
MartinHjelmare, line 10: I don't think we need to lint on all Python versions. I think we should run the tests on all supported Python versions and lint on the minimum supported version.

.github/workflows/test.yml
```diff
@@ -1,31 +1,130 @@
...
     branches: [master]
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  build:
+  info:
```
MartinHjelmare, line 22: I think this workflow is very complex for the size of this project. I'd look into using the builtin caching feature of the setup-python action, and remove the custom virtual environment job and info collection.

MartinHjelmare (changes requested):

PR #734, opened 2023-08-01
Break the `main` module into functions

jbkkd:
For easier reading.

No functional changes in this PR.

### DIFF
--- a/pytradfri/__main__.py
+++ b/pytradfri/__main__.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 import argparse
 import logging
 from pprint import pprint
+from typing import Any
 import uuid
 
 from pytradfri.api.libcoap_api import APIFactory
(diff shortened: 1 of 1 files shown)


pytradfri/__main__.py
```diff
@@ -17,10 +18,11 @@
 
 CONFIG_FILE = "tradfri_standalone_psk.conf"
 
-# pylint: disable=invalid-name, not-an-iterable, unsubscriptable-object
+# pylint: disable=not-an-iterable, unsubscriptable-object
 
 
-if __name__ == "__main__":
+def main() -> None:
```
MartinHjelmare, line 24: This script is meant to be run interactive so that variables are able to be accessed. This won't work if we hide them inside the functions.

MartinHjelmare (changes requested):

PR #735, opened 2023-08-01
Replace 3.10-stretch with 3.11-bullseye

jbkkd:
`stretch` does not exist anymore in the official Python docker repo, so bullseye is used instead.

Also, switching to 3.11, as that is fully supported by this library.


ggravlingen (approved):
Thank you!

PR #815, opened 2024-02-03
Fix reset protocol on error during context shutdown

MartinHjelmare:
- If there's an error during protocol context shutdown we want to make sure the protocol context completes the reset, so that a new protocol context is created the next time the protocol context is needed.
- Example stacktrace (second one): https://github.com/home-assistant/core/issues/105004#issuecomment-1855475069

Needs https://github.com/home-assistant-libs/pytradfri/pull/814


ggravlingen (approved):

MartinHjelmare:
Thanks!

PR #995, opened 2025-02-13
Add missing type annotations

MartinHjelmare:
- Add missing type annotations found by the ANN ruff linter setting group.
- https://docs.astral.sh/ruff/rules/#flake8-annotations-ann


ggravlingen (approved):
🌟

PR #996, opened 2025-02-13
Move blind fixture and complete blind tests

MartinHjelmare:
- The eventual goal is to move the tests of the different device types to the specific modules, eg testing a light device should be done in `test_light.py`.
- Another goal is to move the device fixture data to JSON files instead of having them as Python code. JSON format is how we deserialize the device responses in the API so it makes sense to store the fixtures as JSON.


tests/test_blinds.py
```diff
@@ -24,7 +35,26 @@ def test_device_info_properties(device: Device) -> None:
...
+    assert command.data == {"15015": [{"5536": 50}]}
+
+
+@pytest.mark.parametrize("device", ["blind"], indirect=True)
+def test_trigger_blind(device: Device) -> None:
+    """Test trigger blind."""
+    blind_control = device.blind_control
+    assert blind_control is not None
+    command = blind_control.trigger_blind()
+    assert command.data == {"15015": [{"5523": True}]}
```
MartinHjelmare, line 60: I'm suspecting the boolean should be sent as an integer in this command, since that's how we do it in other device types, but I won't change that code without testing it first with a real device.

tests/test_device.py
```diff
@@ -69,8 +68,6 @@ def device_fixture(request: pytest.FixtureRequest) -> Device:
 
 wall_plugs = (("comment", "device"), [("Wall plug", Device(OUTLET))])
 
-roller_blinds = (("comment", "device"), [("Blind", Device(BLIND))])
```
MartinHjelmare, line 72: This isn't used anywhere.

tests/test_blinds.py
```diff
@@ -1,18 +1,29 @@
...
 
-from .devices import BLIND
+from tests.common import load_fixture
 
 
 @pytest.fixture(name="device")
-def device_fixture() -> Device:
-    """Return Device."""
-    return Device(BLIND)
+def device_fixture(request: pytest.FixtureRequest) -> Device:
```
MartinHjelmare, line 14: This fixture will be moved to a common module when all device types tests have been refactored.

ggravlingen (approved):
Thanks!

